PR #2, opened 2018-12-12
Changed SelectInEnvelopeFunc to select indexed envelopes inside the p…

bryandhill:
…rovided envelope

Rather than the reverse (find all indexed envelopes that have the provided envelope contained inside). Which may be useful, but I don't believe is the intent of this query. Evidenced by SelectInEnvelopeFuncIntersecting not being that way.

---------------

If I'm completely lost, and this was the intent of that function, please let me know and I'll write a pull request to add function to allow a query of indexed envelopes inside the provided one. (Which is the function I need).

Many thanks,
Bryan

### DIFF
--- a/rstar/src/algorithm/selection_functions.rs
+++ b/rstar/src/algorithm/selection_functions.rs
@@ -41,7 +41,7 @@ where
     type ContainmentUnit = T::Envelope;
 
     fn is_contained_in(&self, envelope: &T::Envelope) -> bool {
-        envelope.contains_envelope(&self.envelope)
+        self.envelope.contains_envelope(&envelope)
     }
 }
 


Stoeoef:
Hi, thanks for the PR!

`select_in_envelope` is bugged at the moment :( . 
I'm currently working on a 0.2.0 follow up (0.1.0 was _really_ just to share my progress with the `geo`, I'm surprised seeing people interested in it) that will, among other things, fix this.

I had to change the `SelectionFunction` trait to the following (this cannot be copy & pasted, as it relies on other changes):

```rust
impl<T> SelectionFunction<T> for SelectInEnvelopeFunction<T>
where
    T: RTreeObject,
{
    type ContainmentUnit = T::Envelope;

    fn should_unpack_parent(&self, parent_envelope: &T::Envelope) -> bool {
        self.envelope.intersects(parent_envelope)
    }

    fn should_unpack_leaf(&self, leaf: &T) -> bool {
        self.envelope.contains_envelope(&leaf.envelope())
    }
}
```
which I believe is the correct solution. I really need to unpack every parent that is _intersecting_ the query envelope, as it may contain smaller envelopes inside that are part of the query envelope.
Anyway, those changes are going upstream this weekend, I hope that will be sufficient for you!

bryandhill:
Thanks Stefan! Much appreciated.

Stoeoef:
version 0.1.1 is now up.

PR #3, opened 2018-12-30
Rename feature serde_serialize to serde.

adeschamps:
This follows the recommendation of the serde crate, which is to use
the derive feature rather than depending directly on serde_derive.

This is a breaking change for users of the serde feature, since it
changes the feature name.

### DIFF
--- a/rstar/Cargo.toml
+++ b/rstar/Cargo.toml
@@ -20,13 +20,11 @@ maintenance = { status = "actively-developed" }
 num-traits = "0.2"
 pdqselect = "0.1"
 threadpool = "1.7"
-serde = { version = "1.0", optional = true }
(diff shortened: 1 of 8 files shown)


Stoeoef:
Thanks for the PR!

My serde knowledge is rather out of date as it seems...
I'll give this another look within the next days and will likely merge it then. I'm not too worried about releasing a breaking 0.3.0 instead of 0.2.1, there are some other slightly breaking changes that should be addressed better sooner than later.
Also, your PR indicated some missing words in the documentation (The sentences "Enable the `serde` for ..." are missing the word "feature"), feel free to fix that along the way, otherwise I'll make a commit once I've accepted the PR.

PR #11, opened 2019-07-18
Distance iterator

j-w-shin:
I was looking for a way to retrieve the distances of nearest neighbors but it seems like it needs to be recalculated every time. Is there a way to do this with the current API? Given that the user's `distance_2` function could be arbitrarily complex/expensive, it would be helpful to avoid this recalculation.

I added a new iterator `NearestNeighborDistanceIterator` which emits the underlying distances that preserve ordering in the heap, and made the original `NearestNeighborIterator` into a wrapper for this iterator.

### DIFF
--- a/rstar/src/algorithm/nearest_neighbor.rs
+++ b/rstar/src/algorithm/nearest_neighbor.rs
@@ -42,12 +42,12 @@ where
     }
 }
 
-impl<'a, T> NearestNeighborIterator<'a, T>
+impl<'a, T> NearestNeighborDistanceIterator<'a, T>
 where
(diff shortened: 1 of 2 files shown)


Stoeoef:
Awesome! Seems like the way this should be implemented. Thanks for contributing!

j-w-shin:
Thanks. Could you possibly add this into a new version and publish it to crates.io?

Stoeoef:
Sure, will do!

Stoeoef:
@j-w-shin : 0.5.0 is out now..

PR #14, opened 2019-09-14
Fix nearest neighbor min-max heuristic

rmanoka:
closes Stoeoef/rstar#13


Stoeoef:
LGTM. Thanks for the contribution!

PR #21, opened 2019-11-19
Created a rust.yml file, and ran everything through cargo fmt --all

ckaran:
I did two things in this PR.  First, I setup a workflow that builds and tests after each push.  Second, I ran everything through `cargo fmt --all` and my custom script `trim_rust` that trims the trailing whitespace on all lines in all `.rs` files.  There weren't any semantic changes.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -1,0 +1,15 @@
+name: Rust
+
+on: [push]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v1
+    - name: Build
+      run: cargo build --verbose
+    - name: Run tests
+      run: cargo test --verbose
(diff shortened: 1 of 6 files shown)


Stoeoef:
Thanks for the PR! Is there a specific reason for the switch to GitHub Actions? Are there benefits of using them?

I'm not against that, I just need some more background for why you wanted it as part of this commit. Also, I would argue that it makes sense to remove the Travis CI build file in the same go.

One difference I'm seeing are the test coverage reports that are not part of the git hub pipeline. However, I'm willing to let them go if there is something in favour of GitHub Actions, the test coverage reports weren't that meaningful in

ckaran:
@Stoeoef my apologies for the lateness of this reply, I was on vacation with my family.

Honestly, I'm not too sure why I turned on GitHub actions.  The more I think about it, the less good reason I have for it; the only reason I can think of doing it is due to a brain glitch on my part.  Do you want to edit the PR to remove it?  If not (or you can't), I can redo the PR without GitHub actions.

Stoeoef:
No worries!

I've pushed the commit containing the formatting changes. Thanks for that!

PR #24, opened 2020-01-16
Allow intersections to work with different objects

### LINKED ISSUE #23: Could `intersection_candidates` only require the envelopes to be the same?
First, thank you very much for this crate, it has worked very well for me so far. I like the API and the flexibility, and of course the performance!

I have a case where I want to determine intersection candidates between triangles and tetrahedra. I have two separate RTrees for other purposes (e.g. finding closest triangle/tetrahedron to a point): one for triangles and one for tetrahedra. Unfortunately, `intersection_candidates_with_other_tree` only accepts `&Self` as input, despite the fact that it only works with the envelopes of the two objects, which are the same (`AABB`). Would it be possible to relax `intersection_candidates_with_other_tree` so that it takes any `other` which shares the same envelope?

Andlon:
Only the envelopes are used for determining intersection candidates, so it suffices that the envelopes of the two objects are of the same type.

This would resolve #23. The changes here also have the very fortunate side effect of making sure that `intersection_candidates` always gives tuples `(object1, object2`)`, i.e. so that objects from tree `self` always come first in the returned tuples, and objects from tree 2 come from the `other` tree. Before, it was not possible to distinguish which object came from which tree.

I haven't added any new tests, and I haven't adjusted the documentation, but I could do both. For this, I'll wait for your feedback. I otherwise don't think it should cause any issues in existing code, because the new code is a strict generalization. As I pointed out above, however, it *does* change the order in the tuples returned, but I can't really imagine how any code could be relying on this anyway?

Let me know what you think!

### DIFF
--- a/rstar/src/algorithm/intersection_iterator.rs
+++ b/rstar/src/algorithm/intersection_iterator.rs
@@ -4,18 +4,20 @@ use crate::RTreeNode;
 use crate::RTreeNode::*;
 use crate::RTreeObject;
 
-pub struct IntersectionIterator<'a, T>
+pub struct IntersectionIterator<'a, T, U=T>
 where
     T: RTreeObject,
+    U: RTreeObject,
 {
-    todo_list: Vec<(&'a RTreeNode<T>, &'a RTreeNode<T>)>,
+    todo_list: Vec<(&'a RTreeNode<T>, &'a RTreeNode<U>)>,
 }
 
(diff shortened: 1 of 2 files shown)


Stoeoef:
That's a good observation, we didn't realize before that it's not entirely clear which item comes from which tree.

Thanks for putting this together and pointing that out!

PR #25, opened 2020-02-16
Document crate::point::PointExt

jornvandebeek:
PointExt can be useful when implementing your own RTreeObject or PointDistance. I have documented its functions and exposed it for use.

### DIFF
--- a/rstar/src/lib.rs
+++ b/rstar/src/lib.rs
@@ -43,5 +43,5 @@ pub use crate::envelope::Envelope;
 pub use crate::node::{ParentNode, RTreeNode};
 pub use crate::object::{PointDistance, RTreeObject};
 pub use crate::params::{DefaultParams, InsertionStrategy, RTreeParams};
-pub use crate::point::{Point, RTreeNum};
+pub use crate::point::{Point, PointExt, RTreeNum};
 pub use crate::rtree::RTree;
--- a/rstar/src/lib.rs
+++ b/rstar/src/lib.rs
@@ -43,5 +43,5 @@ pub use crate::envelope::Envelope;
 pub use crate::node::{ParentNode, RTreeNode};
 pub use crate::object::{PointDistance, RTreeObject};
 pub use crate::params::{DefaultParams, InsertionStrategy, RTreeParams};
-pub use crate::point::{Point, PointExt, RTreeNum};
+pub use crate::point::{Point, RTreeNum};
 pub use crate::rtree::RTree;
--- a/rstar/src/point.rs
+++ b/rstar/src/point.rs
@@ -178,11 +178,14 @@ pub trait Point: Copy + Clone + PartialEq + Debug {
 
 impl<T> PointExt for T where T: Point {}
 
+/// Utility functions for Point
 pub trait PointExt: Point {
+    /// Returns a new Point with all components set to zero.
     fn new() -> Self {
         Self::from_value(Zero::zero())
     }
 
+    /// Applies `f` on each pair of components of `self` and `other`.
     fn component_wise(
         &self,
         other: &Self,
@@ -191,6 +194,7 @@ pub trait PointExt: Point {
         Self::generate(|i| f(self.nth(i), other.nth(i)))
     }
 
+    /// Returns whether all pairs of components of `self` and `other` pass test closure `f`. Short circuits if any result is false.
     fn all_component_wise(
         &self,
         other: &Self,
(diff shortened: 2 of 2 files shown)


Stoeoef:
# [Codecov](https://codecov.io/gh/Stoeoef/rstar/pull/25?src=pr&el=h1) Report
> Merging [#25](https://codecov.io/gh/Stoeoef/rstar/pull/25?src=pr&el=desc) into [master](https://codecov.io/gh/Stoeoef/rstar/commit/a3541b0b6452797dd407b7ff11f49ed803a26db7&el=desc) will **increase** coverage by `0.03%`.
> The diff coverage is `n/a`.

[Impacted file tree graph](https://codecov.io/gh/Stoeoef/rstar/pull/25?src=pr&el=tree)

```diff
@@            Coverage Diff             @@
##           master      #25      +/-   ##
==========================================
+ Coverage   98.49%   98.52%   +0.03%     
==========================================
  Files          18       18              
  Lines        1391     1426      +35     
==========================================
+ Hits         1370     1405      +35     
  Misses         21       21              
```


| [Impacted Files](https://codecov.io/gh/Stoeoef/rstar/pull/25?src=pr&el=tree) | Coverage Δ | |
|---|---|---|
| [rstar/src/point.rs](https://codecov.io/gh/Stoeoef/rstar/pull/25/diff?src=pr&el=tree#diff-cnN0YXIvc3JjL3BvaW50LnJz) | `100.00% <0.00%> (ø)` | :arrow_up: |

------

[Continue to review full report at Codecov](https://codecov.io/gh/Stoeoef/rstar/pull/25?src=pr&el=continue).
> **Legend** - [Click here to learn more](https://docs.codecov.io/docs/codecov-delta)
> `Δ = absolute <relative> (impact)`, `ø = not affected`, `? = missing data`
> Powered by [Codecov](https://codecov.io/gh/Stoeoef/rstar/pull/25?src=pr&el=footer). Last update [a3541b0...a149f69](https://codecov.io/gh/Stoeoef/rstar/pull/25?src=pr&el=lastupdated). Read the [comment docs](https://docs.codecov.io/docs/pull-request-comments).

Stoeoef:
Thanks a lot for the PR! And sorry for the delay.

The added documentation is great, I would definitely like to keep that.

Now to the "but": So far, I've hidden `PointExt` on purpose. Making something part of the public API requires to maintain it more thoroughly and increases the pressure to deliver a high quality implementation. That's something I'm not sure `rstar` can provide (at least for the methods of `PointExt`) at the moment, `PointExt` was not designed to be easy to use or ergonomic. Take the `add` and `sub` method for example - while this is fine internally, an external user may (understandably) want a proper `Add` impl. `rstar` avoids the need to provide such by discouraging the use of `Point`.

On the other hand, I do recognize that a more feature rich `Point` type can simplify certain `RTreeObject` and `PointDistance` implementations. There may be a few solutions that are definitely possible and may help:
 - Maybe there is some sort of middle ground by only publishing the most useful functions? On a first glance, `max_point` and `min_point` could be exposed without too much feature creep. Are these already sufficient to address some of the pain points you mentioned?
 - Another solution might be adding support for another lib's point type, e.g. `nalgebra` or `cgmath`. While *direct* support (e.g. adding an `impl Point for cgmath::Point2`) has not worked out so well for me in the past, indirect support via. an interoperability layer could probably be a good way forward. I'm thinking specifically about supporting the [mint crate](https://crates.io/crates/mint).

I hope that these arguments sound reasonable! I would be interested to hear more about the pain points when implementing `RTreeObject` and `PointDistance`. Also, please be not discouraged by this PR not being accepted straight away or may not be able to achieve what you initially have been seeking for. This is not related to the actual code changes (I'd like to keep the changes to `point.rs`!). I hope we can come up with an alternative solution that does not require to extend the public interface.

jornvandebeek:
Thank you for your extensive response. I appreciate that you want to support any exposed interface the best you can.

I have been using PointExt min/max_point, add and sub mostly. Personally I prefer methods over overloaded operators, but that is matter of taste, so if anyone prefers to impl std::ops::Add instead that is fine too. 

Specifically to implement my own RTReeObject + PointDistance, the need for PointExt arised from the `PointDistance::distance_2` fn. It's a oneliner that only becomes like 2 or 3 straightforward lines without PointExt, but I like to copy Rstar's implementation (from Rectangle) verbatim. 
```rust
self.nearest_point(point).sub(point).length_2()
```
One possible solution is for Rstar to implement distance_2 on Point. Then any PointDistance::distance_2 impls can call distance_2 on Point like this, foregoing the PointExt methods: `self.nearest_point(point).distance_2(point)`. 

Then again, I have been using other PointExt methods in other places, so that would be a partial solution for me.

Exposing PointExt in my personal fork is a tiny change that is easy to port forward. I am satisfied with doing that until Rstar has a vector math approach that would be appropriate to expose publicly.

I have edited the PR to just contain the documentation addition to PointExt.

Stoeoef:
> One possible solution is for Rstar to implement distance_2 on Point. Then any PointDistance::distance_2 impls can call distance_2 on Point like this, foregoing the PointExt methods: self.nearest_point(point).distance_2(point).

This should already be implemented, assuming I'm not misunderstanding you. There's a blank implementation `impl <P> PointDistance for P where P: Point` ([source](https://docs.rs/rstar/0.7.1/src/rstar/object.rs.html#208-232)) 

I've opened #28 to address the third party interoperability issues.
Until then, I'll merge the documentation changes - thanks again!

PR #27, opened 2020-03-16
No std support

brenzi:
This PR provides no_std support for rstar crate.
It makes use of (now stable) `alloc`

### DIFF
--- a/rstar/Cargo.toml
+++ b/rstar/Cargo.toml
@@ -17,13 +17,16 @@ codecov = { repository = "Stoeoef/rstar", branch = "master", service = "github" 
 maintenance = { status = "actively-developed" }
 
 [dependencies]
-num-traits = "0.2"
-pdqselect = "0.1"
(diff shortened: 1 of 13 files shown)


urschrei:
@brenzi Sorry about the delay on this. Could you rebase this PR against master, and then we'll see about merging it?

michaelkirk:
I'm not really familiar with no_std practices. 

Once something like this merged, is there best-practice way to ensure we don't regress on no_std support, like something we should add to CI?

urschrei:
Me neither, but it looks like https://github.com/hobofan/cargo-nono might be useful here…

brenzi:
I can rebase towards mid-of-May

brenzi:
As I still haven't found the time (the rebase wasn't entirely straightforward), I have to leave this to you, sorry

wucke13:
Just in case that it is easier, I put together a MVP to make `rstar` compile for no_std over on #83 .

PR #34, opened 2020-05-02
Use should_unpack_leaf in SelectionIteratorMut

therealbnut:
Hey, thanks for the great library :) I noticed that invalid leaves were coming through in my custom selection function, this fixes it.

### DIFF
--- a/rstar/src/algorithm/iterators.rs
+++ b/rstar/src/algorithm/iterators.rs
@@ -84,11 +84,11 @@ where
     Func: SelectionFunction<T>,
 {
     pub fn new(root: &'a mut ParentNode<T>, func: Func) -> Self {
-        let current_nodes = root
-            .children
-            .iter_mut()
-            .filter(|c| func.should_unpack_parent(&c.envelope()))
-            .collect();
+        let should_unpack_node = |node: &&mut RTreeNode<T>| match node {
+            RTreeNode::Leaf(ref t) => func.should_unpack_leaf(t),
+            RTreeNode::Parent(ref data) => func.should_unpack_parent(&data.envelope),
+        };
+        let current_nodes = root.children.iter_mut().filter(should_unpack_node).collect();
         SelectionIteratorMut {
             func,
             current_nodes,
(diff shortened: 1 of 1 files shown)


rstar/src/algorithm/iterators.rs
```diff
@@ -84,11 +84,11 @@ where
     Func: SelectionFunction<T>,
 {
     pub fn new(root: &'a mut ParentNode<T>, func: Func) -> Self {
-        let current_nodes = root
-            .children
-            .iter_mut()
-            .filter(|c| func.should_unpack_parent(&c.envelope()))
```
therealbnut, line 90: This assumed that `should_unpack_leaf` was just `self.should_unpack_parent(&leaf.envelope())`

rstar/src/algorithm/iterators.rs
```diff
@@ -84,11 +84,11 @@ where
...
-            .collect();
+        let should_unpack_node = |node: &&mut RTreeNode<T>| match node {
+            RTreeNode::Leaf(ref t) => func.should_unpack_leaf(t),
+            RTreeNode::Parent(ref data) => func.should_unpack_parent(&data.envelope),
+        };
+        let current_nodes = root.children.iter_mut().filter(should_unpack_node).collect();
```
therealbnut, line 91: This filter isn't actually needed, as they're processed in `next`

rstar/src/algorithm/iterators.rs
```diff
@@ -104,22 +104,21 @@ where
...
+                RTreeNode::Leaf(ref mut t) => {
+                    if self.func.should_unpack_leaf(t) {
+                        return Some(t);
+                    }
+                }
                 RTreeNode::Parent(ref mut data) => {
-                    self.current_nodes.extend(
-                        data.children
-                            .iter_mut()
-                            .filter(|c| func.should_unpack_parent(&c.envelope())),
```
therealbnut, line 115: The filtering here isn't done anymore, so the leaf `envelope` isn't checked if `should_unpack_leaf` doesn't do it. This means `current_nodes` might grow more than it would when checking is eager.

This could be added back, but should be symmetrical with `SelectionIterator`. If that happens `should_unpack_leaf` should probably have docs to say its envelope is already checked, and the trait could default to `true`:
```rust
fn should_unpack_leaf(&self, leaf: &T) -> bool {
    return true;
}
```

rstar/src/algorithm/iterators.rs
Stoeoef: You're absolutely right. Thanks for finding and fixing this!

PR #36, opened 2020-05-14
Fixes a broken doc link for primitives module

Maximkaaa:
Following the link before the fix leads to a `Not Found` error
page.


Stoeoef:
Thanks a lot!
Merging.

PR #37, opened 2020-05-18
RFC: Add support for internal iteration to avoid heap allocations

adamreichold:
We are happily using rstar in ecological simulations of pollinator behavior as a spatial acceleration structure and have noticed at lot of memory allocator activity in our CPU profiles. While there does not seem to be a simple solution to avoid heap allocations for queries, this does provide one possible strategy: Using internal iteration to keep the necessary state on the regular stack. And while this does provide a strictly less versatile and less ergonomic API, it would be sufficient for our use cases.

Please let us know what you think of this approach and whether you would consider it for inclusion upstream. Thank you! (I did implement only the point queries as we only use those for now and the documentation certainly needs extension, especially detailing the increased stack usage.)

### DIFF
--- a/rstar/src/algorithm/iterators.rs
+++ b/rstar/src/algorithm/iterators.rs
@@ -43,10 +43,31 @@ where
     }
 }
 
+pub fn select_nodes<'a, T, Func, Cons>(root: &'a ParentNode<T>, func: &Func, cons: &mut Cons)
+where
+    T: RTreeObject,
+    Func: SelectionFunction<T>,
+    Cons: FnMut(&'a T) -> bool,
+{
+    if func.should_unpack_parent(&root.envelope()) {
+        for node in root.children.iter() {
+            match node {
+                RTreeNode::Leaf(ref t) => {
+                    if func.should_unpack_leaf(t) && cons(t) {
+                        break;
+                    }
+                }
+                RTreeNode::Parent(ref data) => {
+                    select_nodes(data, func, cons);
+                }
+            }
+        }
+    }
+}
+
 impl<'a, T, Func> Iterator for SelectionIterator<'a, T, Func>
 where
     T: RTreeObject,
-
     Func: SelectionFunction<T>,
 {
     type Item = &'a T;
@@ -101,10 +122,10 @@ where
 impl<'a, T, Func> Iterator for SelectionIteratorMut<'a, T, Func>
 where
     T: RTreeObject,
-
     Func: SelectionFunction<T>,
 {
     type Item = &'a mut T;
+
     fn next(&mut self) -> Option<&'a mut T> {
         while let Some(next) = self.current_nodes.pop() {
             match next {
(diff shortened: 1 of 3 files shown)


rstar/src/algorithm/iterators.rs
```diff
@@ -43,10 +43,31 @@ where
...
 
-impl<'a, T, Func> Iterator for SelectionIterator<'a, T, Func>
+pub fn select_nodes<'a, T, Func, Cons>(root: &'a ParentNode<T>, func: &Func, cons: &mut Cons)
 where
     T: RTreeObject,
+    Func: SelectionFunction<T>,
+    Cons: FnMut(&'a T) -> bool,
+{
+    if func.should_unpack_parent(&root.envelope()) {
+        for node in root.children.iter().rev() {
```
adamreichold, line 53: I used `.rev()` to stay exactly compatible with the internal stack-based implementation, but I do wonder whether this is actually necessary?

Stoeoef:
Thanks for opening this!
I haven't considered internal iterators so far. Seeing how they can be used to prevent allocations seems like a good use case indeed.
Another option could be to make `SelectionIterator::current_nodes: Vec<T>` a `SmallVec<T>`. Actually, not having that seems to be an oversight.
I'll compare this version with the internal one. If the `SmallVec` variant already matches the performance of the internal one, I'd prefer that one for maintainability.

Thanks again for pointing this out!

rstar/src/algorithm/iterators.rs
Stoeoef: Nah, the order of elements should be considered random in general. Let's not bother with that.

adamreichold:
> Another option could be to make SelectionIterator::current_nodes: Vec<T> a SmallVec<T>. 

I did consider this, but I have to admit that I do not feel confident about choosing the proper inline capacity here. Should this correspond to `RTreeParams::MAX_SIZE` then?

rstar/src/algorithm/iterators.rs
adamreichold: Done.

Stoeoef:
> > Another option could be to make SelectionIterator::current_nodes: Vec a SmallVec.
> 
> I did consider this, but I have to admit that I do not feel confident about choosing the proper inline capacity here. Should this correspond to `RTreeParams::MAX_SIZE` then?

That's a fair point. The maximum queue length should be in `O(MAX_SIZE * tree_height) = O(MAX_SIZE * log(n))`. Unfortunately, allocating differently sized `SmallVec` will require const generics, which we don't have.

For simplicity, I would suggest a rather arbitrary fixed value, e.g. "32".

Once the queue has a length of 32, allocations might be not the main bottleneck any more - in a tree of that size, a single allocation is probably not noticeable that much. I would expect those trees to be memory bound anyway.
Admittedly, that doesn't work well for larger `MAX_SIZE`, but that parameter isn't modified by most anyway.

That's all a bit theory heavy so far. I'd like to run some benchmarks for comparison and then make a decision. If you like to step in: I'd be interested in how the `SmallVec<[_;32]>` variant fares with your real world scenario.

Also:
> We are happily using rstar in ecological simulations of pollinator 

Awesome! Thanks for sharing that titbit :)

adamreichold:
> If you like to step in: I'd be interested in how the SmallVec<[_;32]> variant fares with your real world scenario.

I am currently measuring the three scenarios but it will take a bit until I have stable numbers...

adamreichold:
> I am currently measuring the three scenarios but it will take a bit until I have stable numbers...

So it looks like better inlining actually favors the `SmallVec` variant, i.e. the results in time steps per second for our current model are:
* `master` does **990** and is rather noisy due to heap fragmentation
* `internal-iterator-to-avoid-heap-allocations` does **1040** with the `select_nodes` function not being inlined and rather hot in the CPU profile
* `use-smallvec-to-avoid-heap-allocations` does **1090** with similarly stable results

It looks like the recursive call overhead is rather prohibitive and I agree with you that the allocation should not be the bottleneck for larger tree, hence I am closing this one in favor of #38

PR #42, opened 2020-06-30
Add cargo fmt to travis build config

Atul9:


.travis.yml
```diff
@@ -4,4 +4,6 @@ rust:
 
 cache: cargo
 script:
-  - cargo build -p rstar-benches && cargo test --all-features -p rstar
\ No newline at end of file
+  - rustup component add rustfmt --toolchain stable-x86_64-unknown-linux-gnu
```
adamreichold, line 7: Is it wise to specify the toolchain here? Doesn't this break using the build matrix if we decide to?

adamreichold:
(While this is very much not a reason to do this, please note that rustfmt is rather limited on this code base due to the large of amount of code being wrapping in macros.)

adamreichold:
We are not using Travis CI any more. Hence I'll close this in favour of adding this functionality to GitHub actions in #119.

PR #57, opened 2021-04-21
Added impl Point for Tuples

nonnominandus:
In my own work I often use tuples to represent points (as the dimensionality is fixed by definition) so I was lacking the option to easily work with tuples with the rtree. Of course I can make my own wrapper functions to make it workable. But it felt like there was no theoretical limitation holding Tuple types back from implementing Point. I tried really hard to create a macro generating the tuple implementations, but ultimately failed (it proved to be hard to generate the Tuple type), so I copied all implementations for all lengths. This is obviously not a very elegant solution feel free to clean it up, but it works in this way and it is proven to work by the included unit test. I also added an example of how this works in the documentation, feel free to leave it out or expand it if you feel that is necessary.


adamreichold:
Please have a look at https://doc.rust-lang.org/stable/src/core/tuple.rs.html#7 or better yet https://doc.rust-lang.org/stable/src/core/hash/mod.rs.html#612 on how to handle the repetitive impls using macros.

adamreichold:
From a design point of view, I am not sure I see a benefit of using tuples compared to fixed size arrays as the traits here enforce homogenous types anyway, so the primary difference between tuples and arrays is not useful. Furthermore, supporting fixed size arrays will become significantly easier with const generics (even the MVP that recently hit stable).

Is there anything that can be done using tuples which is not possible using arrays?

rstar/src/point.rs
```diff
@@ -343,3 +345,387 @@ implement_point_for_array!(0, 1, 2, 3, 4, 5);
 implement_point_for_array!(0, 1, 2, 3, 4, 5, 6);
 implement_point_for_array!(0, 1, 2, 3, 4, 5, 6, 7);
 implement_point_for_array!(0, 1, 2, 3, 4, 5, 6, 7, 8);
+
+// The following code screams to be implemented with a macro, but this is not
+// possible with a simple declarative, a procedural macro would be a possibility
```
adamreichold, line 350: I don't think this is a accurate, e.g.
```rust
macro_rules! fixed_type {
    ($expr:expr, $type:ty) => { $type };
}

macro_rules! impl_point_for_tuple {
    ($($index:expr => $name:ident),+) => {
        impl<S> Point for ($(fixed_type!($index, S),)+)
        where
            S: RTreeNum
        {
            type Scalar = S;

            const DIMENSIONS: usize = count_exprs!($($index),*);

            fn generate(generator: impl Fn(usize) -> S) -> Self {
                ($(generator($index),)+)
            }

            #[inline]
            fn nth(&self, index: usize) -> Self::Scalar {
                let ($($name,)+) = self;

                match index {
                    $($index => *$name,)+
                    _ => unreachable!(),
                }
            }

            #[inline]
            fn nth_mut(&mut self, index: usize) -> &mut Self::Scalar {
                let ($($name,)+) = self;

                match index {
                    $($index => $name,)+
                    _ => unreachable!(),
                }
            }
        }
    };
}

impl_point_for_tuple!(0 => a);
impl_point_for_tuple!(0 => a, 1 => b);
impl_point_for_tuple!(0 => a, 1 => b, 2 => c);
impl_point_for_tuple!(0 => a, 1 => b, 2 => c, 3 => d);
impl_point_for_tuple!(0 => a, 1 => b, 2 => c, 3 => d, 4 => e);
impl_point_for_tuple!(0 => a, 1 => b, 2 => c, 3 => d, 4 => e, 5 => f);
impl_point_for_tuple!(0 => a, 1 => b, 2 => c, 3 => d, 4 => e, 5 => f, 6 => g);
impl_point_for_tuple!(0 => a, 1 => b, 2 => c, 3 => d, 4 => e, 5 => f, 6 => g, 7 => h);
impl_point_for_tuple!(0 => a, 1 => b, 2 => c, 3 => d, 4 => e, 5 => f, 6 => g, 7 => h, 8 => i);
impl_point_for_tuple!(0 => a, 1 => b, 2 => c, 3 => d, 4 => e, 5 => f, 6 => g, 7 => h, 8 => i, 9 => j);
```

nonnominandus:
That seems like a great way to solve the problem! I am not that accustomed to macro writing yet so it proved to be too difficult for me yesterday. With your permission I pushed your macro implementation to the PR.

> Is there anything that can be done using tuples which is not possible using arrays?

I would turn the question around: is there any reason for not supporting homogenous tuples? Because they essentially contain the same information as fixed size arrays. I would argue that an uninitiated programmer could expect tuples to also work, I certainly did. Coming from other languages, missing the fixed size arrays, I tend to work with tuples to constrain the dimensionality of a point. Also I like that you cannot have dynamic access to tuple members (indexing) which means that in your source code have to be clear which member (and so which dimension) you expect to handle, this makes it clear to me what is happening.

One reason I see for not including this is the fact that it violates the rules of the Tuple definition, in that it allows for dynamic access to tuple members.

nonnominandus:
If possible I would like this PR to be included in the next update, that would make the next update of my crate a little bit more sleek and remove the need for very fast deprecation of some included functions. Is there something I can do to help?

I am not sure who can help me with this but maybe @urschrei can help me?

nonnominandus:
Hey there, is there any news on the plans for inclusion of this PR? I would like for it to be merged so that I can update my own crate. Is there anything I can do to helps its adoption?

As you were active @rmanoka and added some very good features recently maybe you could shed some light on this?

rmanoka:
I'm on the fence on including it.  Libraries like [nalgebra](https://docs.rs/nalgebra/0.29.0/nalgebra/base/struct.Matrix.html) and [cgmath](https://docs.rs/cgmath/0.18.0/cgmath/struct.Matrix2.html) seem to support only arrays and not tuples;  on the other hand, [geo](https://docs.rs/geo/0.18.0/geo/struct.Coordinate.html) does support conversion from a tuple.

@urschrei @Stoeoef @adamreichold any votes?

adamreichold:
Personally, I prefer using tuples for small dimensions and it seems a nice API courtesy for users of this crate. The implementation does not seem more involved than the array-based one either.

The main downside that I see is that things will become significantly simpler and more general for arrays with const generics but that will then not apply to tuples, i.e. we will always have to keep the macro tricks around even if the array part becomes simpler due to const generics.

rmanoka:
Sounds good.  Will give it a couple of days, and merge if no one objects.  We could try to push for a release in the next week or two.

nonnominandus:
Great! Thanks all!

rmanoka:
@nonnominandus could you add a line in CHANGES.md referencing this PR?

nonnominandus:
I hope this worked out as intended and I did not cause any git hell. I added a line to `rstar/CHANGELOG.md` explaining the changes.

rstar/src/point.rs
```diff
@@ -344,3 +346,90 @@ implement_point_for_array!(0, 1, 2, 3, 4, 5);
...
+                ($(generator($index),)+)
+            }
+
+            #[inline]
+            fn nth(&self, index: usize) -> Self::Scalar {
+                let ($($name,)+) = self;
+
+                match index {
+                    $($index => *$name,)+
+                    _ => unreachable!(),
```
michaelkirk, line 376: little nit: specific error text would be helpful in case we ever hit it.

```suggestion
                    _ => unreachable!("index {} out of bounds for tuple", index),
```

rstar/src/point.rs
```diff
@@ -344,3 +346,90 @@ implement_point_for_array!(0, 1, 2, 3, 4, 5);
...
+                }
+            }
+
+            #[inline]
+            fn nth_mut(&mut self, index: usize) -> &mut Self::Scalar {
+                let ($($name,)+) = self;
+
+                match index {
+                    $($index => $name,)+
+                    _ => unreachable!(),
```
michaelkirk, line 386:
```suggestion
                    _ => unreachable!("index {} out of bounds for tuple", index),
```

nonnominandus:
@michaelkirk I do agree I added your suggestions.

michaelkirk:
bors try

michaelkirk:
Currently rstart compiles on rust at least as far back as 1.50 (see https://github.com/georust/rstar/runs/3389089421?check_suite_focus=true), but it seems like with your changes that'd no longer be true.

What versions are you testing your changes one? Is there a reasonable way to continue to support back to 1.50 or do we need to look at bumping our MSRV (and what to?).

rstar/src/point.rs
```diff
@@ -344,3 +346,90 @@ implement_point_for_array!(0, 1, 2, 3, 4, 5);
...
+    ($($index:expr => $name:ident),+) => {
+        impl<S> Point for ($(fixed_type!($index, S),)+)
+        where
+            S: RTreeNum
+        {
+            type Scalar = S;
+
+            const DIMENSIONS: usize = count_exprs!($($index),*);
+
+            fn generate(generator: impl Fn(usize) -> S) -> Self {
```
adamreichold, line 366: I think this should indeed be
```rust
fn generate(generator: impl FnMut(usize) -> S) -> Self;
```
as in the [trait definition](https://github.com/georust/rstar/blob/62f44f3f649eda46efe0c6a96c90b0e8a9e7586d/rstar/src/point.rs#L169) to fix the build errors.

nonnominandus:
bors try

nonnominandus:
Oh sorry I was unsure how to get bors to run.

nonnominandus:
After one additional tiny change all tests from the github workflow are successful.

michaelkirk:
bors try

nonnominandus:
Thanks everyone! Happy that my first ever code contribution to a crate is merged.
Special thanks to @adamreichold for helping me with the macro.

PR #60, opened 2021-04-27
Fix rstar-demo for aarch64 macos

michaelkirk:
- [-] I added an entry to `CHANGES.md` if knowledge of this change could be valuable to users.
---

The current demo wasn't compiling on my machine - failing due to an old downstream dep of kiss3d (winit). Fixed by the update.

Screen Shot 2021-04-27 at 10 06 19 AM

### DIFF
--- a/rstar-demo/Cargo.toml
+++ b/rstar-demo/Cargo.toml
@@ -6,7 +6,12 @@ edition = "2018"
 
 [dependencies]
 rand = "0.7"
-kiss3d = "0.31"
-nalgebra = "0.26"
+# using a slightly older version of kiss3d/nalgebra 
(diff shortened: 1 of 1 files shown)


michaelkirk:
... investigating ci error.

michaelkirk:
> ... investigating ci error.

The latest-and-greatest nalgebra crate doesn't work on rust pre-1.51, which was just release about a month ago.

Rather than answer hard questions about how long we should endeavor to support old versions of rust, since it's just a demo-crate, I've opted to peg it to a slightly-less-than-latest release for now.

urschrei (approved):

michaelkirk:
bors r=urschrei

PR #61, opened 2021-04-27
Clippy, rustfmt, Cargo metadata updates

urschrei:
- [x] I added an entry to `CHANGES.md` if knowledge of this change could be valuable to users.
---

### DIFF
--- a/rstar/Cargo.toml
+++ b/rstar/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "rstar"
 version = "0.8.3"
-authors = ["Stefan Altmayer <stoeoef@gmail.com>"]
+authors = ["Stefan Altmayer <stoeoef@gmail.com>", "The Georust Developers <mods@georust.org>"]
 description = "R*-tree library for the rust ecosystem"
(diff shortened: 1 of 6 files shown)


rstar/Cargo.toml
```diff
@@ -1,19 +1,17 @@
 [package]
 name = "rstar"
-version = "0.8.3"
-authors = ["Stefan Altmayer <stoeoef@gmail.com>"]
+version = "0.8.4"
```
michaelkirk, line 3: No big deal, but I usually don't bump the version until just before the release;

1. we don't know what other (breaking) changes might go in before release
2. less to keep track of so you don't have to double-check every time you make a PR if you should be bumping the version

Is there a benefit to bumping now?

michaelkirk (approved):
LGTM - just some "take it or leave it" feedback.

rstar/Cargo.toml
urschrei: No, it was purely a "we'll be bumping anyway" thing. Will revert.

urschrei:
bors r=michaelkirk

PR #67, opened 2021-05-06
Add a `SelectionWithDataIterator` trait that allows `should_unpack_leaf` to return `Option<Data>`

expenses:
---

This is useful for things like ray intersections, where you want to return the time of impact in order to find the closest object.

### DIFF
--- a/rstar/CHANGELOG.md
+++ b/rstar/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## Added
 - `RTree::nearest_neighbors` method based on
   [spade crate's implementation](https://github.com/Stoeoef/spade)
-- Added a `SelectionWithDataIterator` trait.
+- Added `RTree::locate_with_selection_function_with_data` and a `SelectionWithDataIterator` trait to go with it.
 
 ## Changed
 - BREAKING: `Point::generate` function now accepts a `impl FnMut`. Custom implementations of `Point` must change to
--- a/rstar/CHANGELOG.md
+++ b/rstar/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Added
 - `RTree::nearest_neighbors` method based on
   [spade crate's implementation](https://github.com/Stoeoef/spade)
+- Added a `SelectionWithDataIterator` trait.
 
 ## Changed
 - BREAKING: `Point::generate` function now accepts a `impl FnMut`. Custom implementations of `Point` must change to
@@ -35,7 +36,7 @@
 ## Changed:
  - Deprecated `RTree::nearest_neighbor_iter_with_distance`. The name is misleading, use `RTree::nearest_neighbor_iter_with_distance_2` instead.
  - Some performance improvements, see #38 and #35
- 
+
 ## Added
  - Added `nearest_neighbor_iter_with_distance_2` #31
 
--- a/rstar/CHANGELOG.md
+++ b/rstar/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Added
 - Add `RTree::drain_*` methods to remove and drain selected items. ([PR](https://github.com/georust/rstar/pull/77))
+- Added `RTree::locate_with_selection_function_with_data` and a `SelectionWithDataIterator` trait to go with it.
 
 ## Changed
 - Expose all iterator types in `crate::iterators` module ([PR](https://github.com/georust/rstar/pull/77))
@@ -16,7 +17,6 @@
 ## Added
 - `RTree::nearest_neighbors` method based on
   [spade crate's implementation](https://github.com/Stoeoef/spade)
-- Added `RTree::locate_with_selection_function_with_data` and a `SelectionWithDataIterator` trait to go with it.
 
 ## Changed
 - Fix floating point inconsistency in `min_max_dist_2` ([PR](https://github.com/georust/rstar/pull/40)).
--- a/rstar/src/algorithm/iterators.rs
+++ b/rstar/src/algorithm/iterators.rs
@@ -127,6 +127,63 @@ where
     }
 }
 
+
+pub struct SelectionWithDataIteratorMut<'a, T, D, Func>
+where
+    T: RTreeObject + 'a,
+    Func: SelectionFunctionWithData<T, D>,
+{
+    func: Func,
+    current_nodes: SmallVec<[&'a mut RTreeNode<T>; 32]>,
+    _phantom: std::marker::PhantomData<D>,
+}
+
+impl<'a, T, D, Func> SelectionWithDataIteratorMut<'a, T, D, Func>
+where
+    T: RTreeObject,
+    Func: SelectionFunctionWithData<T, D>,
+{
+    pub fn new(root: &'a mut ParentNode<T>, func: Func) -> Self {
+        let current_nodes = if func.should_unpack_parent(&root.envelope()) {
+            root.children.iter_mut().collect()
+        } else {
+            SmallVec::new()
+        };
+
+        SelectionWithDataIteratorMut {
+            func,
+            current_nodes,
+            _phantom: std::marker::PhantomData,
+        }
+    }
+}
+
+impl<'a, T, D, Func> Iterator for SelectionWithDataIteratorMut<'a, T, D, Func>
+where
+    T: RTreeObject,
+    Func: SelectionFunctionWithData<T, D>,
+{
+    type Item = (&'a mut T, D);
+
+    fn next(&mut self) -> Option<(&'a mut T, D)> {
+        while let Some(next) = self.current_nodes.pop() {
+            match next {
+                RTreeNode::Leaf(ref mut t) => {
+                    if let Some(data) = self.func.should_unpack_leaf(t) {
+                        return Some((t, data));
+                    }
+                }
+                RTreeNode::Parent(ref mut data) => {
+                    if self.func.should_unpack_parent(&data.envelope) {
+                        self.current_nodes.extend(&mut data.children);
+                    }
+                }
+            }
+        }
+        None
+    }
+}
+
 pub struct SelectionIteratorMut<'a, T, Func>
 where
     T: RTreeObject + 'a,
--- a/rstar/src/algorithm/iterators.rs
+++ b/rstar/src/algorithm/iterators.rs
@@ -71,6 +71,62 @@ where
     }
 }
 
+pub struct SelectionWithDataIterator<'a, T, D, Func>
+where
+    T: RTreeObject + 'a,
+    Func: SelectionFunctionWithData<T, D>,
+{
+    func: Func,
+    current_nodes: SmallVec<[&'a RTreeNode<T>; 24]>,
+    _phantom: std::marker::PhantomData<D>,
+}
+
+impl<'a, T, D, Func> SelectionWithDataIterator<'a, T, D, Func>
+where
+    T: RTreeObject,
+    Func: SelectionFunctionWithData<T, D>,
+{
+    pub fn new(root: &'a ParentNode<T>, func: Func) -> Self {
+        let current_nodes = if func.should_unpack_parent(&root.envelope()) {
+            root.children.iter().collect()
+        } else {
+            SmallVec::new()
+        };
+
+        SelectionWithDataIterator {
+            func,
+            current_nodes,
+            _phantom: std::marker::PhantomData,
+        }
+    }
+}
+
+impl<'a, T, D, Func> Iterator for SelectionWithDataIterator<'a, T, D, Func>
+where
+    T: RTreeObject,
+    Func: SelectionFunctionWithData<T, D>,
+{
+    type Item = (&'a T, D);
+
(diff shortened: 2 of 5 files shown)


rstar/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## Added
 - `RTree::nearest_neighbors` method based on
   [spade crate's implementation](https://github.com/Stoeoef/spade)
+- Added a `SelectionWithDataIterator` trait.
```
adamreichold, line 6: I think the trait is `SelectionFunctionWithData`, isn't it? I would suggest referencing the primary API entry point `` here as well to help people who read the change log and want to try this out.

rstar/src/algorithm/iterators.rs
```diff
@@ -71,6 +71,62 @@ where
...
+    T: RTreeObject,
+    Func: SelectionFunctionWithData<T, D>,
+{
+    type Item = (&'a T, D);
+
+    fn next(&mut self) -> Option<(&'a T, D)> {
+        while let Some(next) = self.current_nodes.pop() {
+            match next {
+                RTreeNode::Leaf(ref t) => {
+                    if let Some(data) = self.func.should_unpack_leaf(t) {
```
adamreichold, line 115: I wonder if we could reduce the amount of duplicate code if this would modify the existing `SelectIterator(Mut)` implementations and to always use `SelectionFunctionWithData`, just with a zero sized type like `()`?

Assuming a blanket impl of `SelectionFunctionWithData<T, ()>` for `SelectionFunction<T>`  is able to bridge between the two public API in a way that is backwards compatible?

(This would also provide a mutable variant of this query which I would suggest adding for completeness if above approach is not tenable.)

expenses: I tried that approach first, but it required a ton of changes and I didn't really find it tenable. I'll add the mutable variant though.

adamreichold: I don't mean to make this complicated but could expand on which intrusive changes were required?

I just tried the diff
```diff
diff --git a/rstar/src/algorithm/selection_functions.rs b/rstar/src/algorithm/selection_functions.rs
index 3aaeda5..9634b3f 100644
--- a/rstar/src/algorithm/selection_functions.rs
+++ b/rstar/src/algorithm/selection_functions.rs
@@ -53,6 +53,20 @@ where
     fn should_unpack_leaf(&self, _leaf: &T) -> Option<D>;
 }
 
+impl<F, T> SelectionFunctionWithData<T, ()> for F
+where
+    F: SelectionFunction<T>,
+    T: RTreeObject,
+{
+    fn should_unpack_parent(&self, envelope: &T::Envelope) -> bool {
+        SelectionFunction::should_unpack_parent(self, envelope)
+    }
+
+    fn should_unpack_leaf(&self, leaf: &T) -> Option<()> {
+        SelectionFunction::should_unpack_leaf(self, leaf).then(|| ())
+    }
+}
+
 pub struct SelectInEnvelopeFunction<T>
 where
     T: RTreeObject,
diff --git a/rstar/src/rtree.rs b/rstar/src/rtree.rs
index 726f4a6..a02a33d 100644
--- a/rstar/src/rtree.rs
+++ b/rstar/src/rtree.rs
@@ -376,7 +376,7 @@ where
         &self,
         selection_function: S,
     ) -> impl Iterator<Item = &T> {
-        SelectionIterator::new(&self.root, selection_function)
+        SelectionWithDataIterator::new(&self.root, selection_function).map(|(item, ())| item)
     }
 
     /// Mutable variant of [`locate_with_selection_function`](#method.locate_with_selection_function).
@@ -384,7 +384,7 @@ where
         &mut self,
         selection_function: S,
     ) -> impl Iterator<Item = &mut T> {
-        SelectionIteratorMut::new(&mut self.root, selection_function)
+        SelectionWithDataIteratorMut::new(&mut self.root, selection_function).map(|(item, ())| item)
     }
 
     /// Similar to [`locate_with_selection_function`](#method.locate_with_selection_function) but with
```
on top your branch and it does seem to build and test.

Is the problem the other explicitly named iterators based on `SelectionIterator{,Mut}`? Maybe writing a small named adaptor like `IgnoreData` that essentially does `.map(|(item, ())| item)` does so that those could be rewritten as e.g.
```rust
pub type LocateAllAtPoint<'a, T> = IgnoreData<SelectionIteratorWithData<'a, T, (), SelectAtPointFunction<T>>>;
```
could resolve this? Or to keep the public iterator type the same, reimplementing it as
```rust
pub struct SelectionIterator<'a, T, Func>(pub(crate) SelectionWithDataIterator<'a, T, (), Func>)
where
    T: RTreeObject + 'a,
    Func: SelectionFunction<T>;

impl<'a, T, Func> Iterator for SelectionIterator<'a, T, Func>
where
    T: RTreeObject,
    Func: SelectionFunction<T>,
{
    type Item = &'a T;

    fn next(&mut self) -> Option<&'a T> {
        self.0.next().map(|(item, ())| item)
    }
}
```
might be another option?

expenses: Oh sorry, I think I misread. I'll try those changes.

adamreichold: (Actually I think it would be simplest if we would demote all iterators returned by `RTree` methods to `impl Iterator<..>` but that should be a maintainer decision due to the implications for downstream code. We currently make those implications for some methods, so we should probably at least strive to be consistent, i.e. provide named types for all or none of the returns. cc @Stoeoef @urschrei)

urschrei: I don't have an immediate answer here – I'm in favour of consistency, though, and I'm fine with a breaking change (in my role as someone who's mostly facilitating maintenance of the crate, as opposed to developing it) if it's warranted. @rmanoka what do you think?

OvermindDL1: It should be named types for all.  Hiding the type means we can't reference the iterator in other structures to be consumed elsewhere without incurring a dynamic dispatch or boxing it up, and on a performance focused library like this considering how it's used, that's important.

rmanoka: @urschrei Sorry, I seem to have missed this comment.  I agree with @OvermindDL1 :  while we could expect the users to use generics if they wish to store the iterator, it makes such usage quite complicated.  Also, std. library seems to expose [most](https://doc.rust-lang.org/std/slice/struct.Iter.html) of [their](https://doc.rust-lang.org/std/iter/struct.Enumerate.html) iterator structures.

rstar/src/algorithm/selection_functions.rs
```diff
@@ -37,6 +37,22 @@ where
     }
 }
 
+/// Similar to [`SelectionFunction`](trait.SelectionFunction.html) but additional data can be returned by
+/// `should_unpack_leaf`.
+pub trait SelectionFunctionWithData<T, D>
+where
+    T: RTreeObject,
+{
```
rmanoka, line 45: Would it be simpler to simply pass the `RTreeNode` directly?  The assumption is that root is implicitly unpacked as it's just a `Vec<RTreeNode>`.  Then the trait just needs one method and we don't have to force that the decision is made based on envelope alone.

@expenses @adamreichold @urschrei

adamreichold: This does sound sensible on a first reading, but I think such a change would be orthogonal to what is proposed here. (It would apply to the original `SelectionFunction` trait just as well.) Hence I would suggest not making this PR any more complicated than it already is and tackling this separately.

rmanoka: Apologies, I didn't realize this is an extension of the existing `SelectionFunction` trait.  But, now I'm more confused.  Given the data `D` is being computed during the `should_unpack_leaf` call,  can't it just be computed during the iteration using the vanilla `locate_with_selection_function`?   What's the extra use-case being handled by this new trait?  Sorry if I'm missing something obvious here.

adamreichold: I'd say it is an "API courtesy". While you could have a mutable binding initialised to `None` and set to `Some` within the selection function before it returns true and then use this in downstream code, that is somewhat contrived, possibly less efficient and really hard to read when iterator adaptors are chained onto the selection method call.

rstar/src/algorithm/iterators.rs
Stoeoef: @urschrei I also agree that iterators should be named. `impl Iterator` in return position is rarely useful for (most) libraries IMO.

rstar/src/algorithm/selection_functions.rs
Stoeoef: Is there a good example which could describe this use case further? I'm wondering how this could be documented, partially because I still don't understand what this could be used for.

Also, if this is about information calculated *within* the selection function, shouldn't the signatures of `SelectionFunctionWithData` rather be `&mut self` (instead of `&self`)?

adamreichold: The original poster mentioned

> This is useful for things like ray intersections, where you want to return the time of impact in order to find the closest object.

in the cover letter.

> Also, if this is about information calculated within the selection function, shouldn't the signatures of SelectionFunctionWithData rather be &mut self (instead of &self)?

I don't think that computing a new value and returning that necessarily requires modifying state just as computing a `bool` indicating whether to unpack or not does not require it.

But then if the implementation can provide an exclusive reference, that would provide a strictly more general API by offering both choices of `&self` and `&mut self` to downstream code and would thereby seem to be preferable. But this line of reasoning would just as well apply to the existing `SelectionFunction` trait again.

expenses:
I'm going to try and get this PR merged over the next few days. A consensus on what the change should look like would be helpful.

rmanoka:
@expenses please provide an example explaining the intended usage of this API.

rstar/src/algorithm/iterators.rs
expenses: So I'm looking into your comments again now, @adamreichold. I'm kinda of the opinion that while we could reduce code duplication here by having everything use a `SelectionWithDataIterator`, it does make everything a bit more complicated. I'm not totally opposed to it, but I think there's an argument to be made for keeping things simple and duplicated as well.

expenses:
@rmanoka happy to do this, I'm not sure where in the repo I should put this example though.

What I'm doing is that I have a triangle type like so:

```rust
pub struct Triangle {
    pub a: Vec3,
    pub b: Vec3,
    pub c: Vec3,
}

impl rstar::RTreeObject for Triangle {
    type Envelope = rstar::AABB<[f32; 3]>;
    ...
```

along with an an `RTree` as part of the data for a 3d model:

```rust
pub struct Model {
    pub acceleration_tree: rstar::RTree<Triangle>,
    ...
}
```

Then as a selection function I have a `Ray`:

```rust
pub struct Ray {
    pub origin: Vec3,
    pub direction: Vec3,

}

impl rstar::SelectionFunctionWithData<Triangle, f32> for Ray {
    fn should_unpack_parent(&self, envelope: &rstar::AABB<[f32; 3]>) -> bool {
        let bounding_box = BoundingBox::new(envelope.lower().into(), envelope.upper().into());
        self.bounding_box_intersection(bounding_box).is_some()
    }

    fn should_unpack_leaf(&self, triangle: &Triangle) -> Option<f32> {
        self.triangle_intersection(triangle)
    }
}
```

The `f32` here being the time-to-impact.

This means that I can find out which model the ray hits first:


```rust
let ray = ...

let first_hit = models.iter()
    .filter_map(|model|
        model.acceleration_tree.locate_with_selection_function_with_data(ray)
            .map(|time_to_impact| (model.id, time_to_impact))
    )
    .min_by(|(_, toi_a), (_, toi_b)| toi_a.partial_cmp(toi_b).unwrap_or(std::cmp::Ordering::Equal))
    .map(|(model_id, _) model_id);
```

The above was written in this comment and won't compile, but you get the idea.

rmanoka:
@expenses Thanks for the example;  that makes the use-case clear.  As discussed in [this thread](https://github.com/georust/rstar/pull/67#discussion_r681964468), I wonder if this could be done with the `locate_with_selection_function`:

```rust
let time_to_impact = acceleration_tree.locate_with_selection_function(&ray)
    .map(|triangle| ray.triangle_intersection(triangle));
```

You would have to adjust the impl `SelectionFunction` to work on reference: `&Ray`; the `should_unpack_parent` would remain the same, and `should_unpack_leaf` could always return `true` (which is the default impl. anyway).  Also, if `Ray::triangle_intersection` returns `Option<f64>`, then you would have to use `filter_map`.

It seems like extra functionality to return data in `unpack_leaf` can instead be done via a `map` or `filter_map` on the iterator obtained from the existing `locate_with_selection_function`.  Am I missing something?

expenses:
> It seems like extra functionality to return data in unpack_leaf can instead be done via a map or filter_map on the iterator obtained from the existing locate_with_selection_function. Am I missing something?

Nope, don't think so! That's indeed something that could work. I think I'll close this PR then.

PR #73, opened 2021-07-31
Added LineWithData and RectangleWithData

jblindsay:
---
I have added a data variants of the Line and Rectangle primitives, in order to allow the association of attribute data.

### DIFF
--- a/rstar/src/primitives/line_with_data.rs
+++ b/rstar/src/primitives/line_with_data.rs
@@ -1,0 +1,145 @@
+use crate::aabb::AABB;
+use crate::envelope::Envelope;
+use crate::object::PointDistance;
+use crate::object::RTreeObject;
+use crate::point::{Point, PointExt};
+use num_traits::{One, Zero};
+
+/// A line defined by a start and and end point with associated data.
+///
+/// This struct can be inserted directly into an r-tree.
+/// # Type parameters
+/// `T`: The line's data type.
+/// `P`: The line's [Point] type.
+///
+/// # Example
+/// ```
+/// use rstar::primitives::LineWithData;
+/// use rstar::{RTree, RTreeObject};
+///
+/// let line_1 = LineWithData::new(1usize, [0.0, 0.0], [1.0, 1.0]);
+/// let line_2 = LineWithData::new(2usize, [0.0, 0.0], [-1.0, 1.0]);
+/// let tree = RTree::bulk_load(vec![line_1, line_2]);
+///
+/// assert!(tree.contains(&line_1));
+/// ```
+#[derive(Debug, Clone, Copy, Eq, PartialEq, Ord, PartialOrd)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct LineWithData<T, P>
+where
+    P: Point,
+{
+    /// The line's data.
+    pub data: T,
+    /// The line's start point.
+    pub from: P,
+    /// The line's end point.
+    pub to: P,
+}
+
+impl<T, P> LineWithData<T, P>
+where
+    P: Point,
+{
+    /// Creates a new line between two points.
+    pub fn new(data: T, from: P, to: P) -> Self {
+        LineWithData { data, from, to }
+    }
+}
+
+impl<T, P> RTreeObject for LineWithData<T, P>
+where
+    P: Point,
+{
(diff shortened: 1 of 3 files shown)


adamreichold:
I wonder if we can reduce the amount of redundant code using that there should be no runtime cost to defining
```rust
type Line<P> = LineWidthData<(), P>;
type Rectangle<P> = RectangleWithData<(); P>;
```
or outright reducing the API surface by using
```rust
struct Line<D=(), P> { ... } where P: Point;
```
i.e. always having that data field present but defaulting to a ZST.

rmanoka:
@jblindsay Would it not be ergonomic to have a generic container RTreeObjectWithData(pub R, pub T) where R: RTreeObject, and T is an arbitrary type?  Then, user could use whichever geom type as needed.

jblindsay:
@adamreichold and @rmanoka Both of your suggestions are great and would likely be better overall solutions. I chose this approach because it had minimal impact on the code base, with no breaking changes. However, if you'd like to move forward with one of these more generic solutions instead, I say go for it!

adamreichold:
I think the `RTreeObjectWithData` approach is the most preferable among the ideas proposed here as it is strictly more general and avoids a boilerplate ZST if there is no associated data. Since `impl RTreeObject for P where P: Point`, we could drop `PointWithData` entirely then.

rmanoka:
@jblindsay Just sketching a quick container to assess usability.

```rust
use rstar::RTreeObject;

#[derive(Debug, Clone, Copy, PartialEq, Eq, Ord, PartialOrd)]
pub struct Container<T, R>(pub T, pub R);

impl<T, R: RTreeObject> RTreeObject for Container<T, R> {
    type Envelope = R::Envelope;

        self.1.envelope()
    }
}
```

In the associated issue, you've mentioned it wasn't easy to link metadata with the geometry.  Does the above container for your use-cases?  The usage I'm imagining is something like:

``` rust
use rstar::{RTree, primitives::Line};

#[test]
fn container_in_rtree() {
    let line_1 = Container((), Line::new([0.0, 0.0], [1.0, 1.0]));
    let line_2 = Container((), Line::new([0.0, 0.0], [-1.0, 1.0]));
    let tree = RTree::bulk_load(vec![line_1, line_2]);
    assert!(tree.contains(&line_1));
}
```

I personally think we should spend more time before adding many container types to the crate as it increase the API surface, and hence the maintenance burden.  The above code is fairly succinct, so we could add after some feedback on usage.  Further, it can easily be added outside the crate, so I think it may be prudent to hear more feedback / use-cases before adding it.

adamreichold:
@rmanoka If such a type is included, I would suggest not deriving `PartialEq` etc. so that it is easier for downstream users to implement these traits with the semantics they want without having to add a newtype wrapper (assuming their crate owns the `T` type).

rmanoka:
@adamreichold If the type is included in this crate, how will they impl standard traits on this type?  Won't rust prevent it?

adamreichold:
IIRC, if the `T` type parameter in the above type definition is owned by their crate, then coherence rules will allow impls like
```rust
impl ForeignTrait for ForeignType<LocalType> { ... }
```

adamreichold:
Ah sorry, I did not remember correctly. This works if the type parameter is part of the trait, not of a generic type... So the point is moot.

rmanoka:
@adamreichold Ya, I'm not an expert on the rust coherence rules either;  I figured it's useful to provide the `RTree::contains` method which seems to require `PartialEq`.  

There are still a few more questions to consider about usability / correctness.  For instance, `PointWithData` seems to make the point object private and read-only after initialization.  The source comments say it is to prevent editing it while in the RTree.   While understandable, the `RTree::iter_mut` and similar explicitly mention that it is a logic error to modify the geometry in place, so it is expected that the users don't violate this.  I just find it inconsistent that `Line` would allow the geom to be edited, but not `Container<T, Line>` but if it prevents common errors, it's worth adding some getters and making it private I suppose.

urschrei:
Closed in favour of #74, which has been released as [rstar 0.9.1](https://docs.rs/rstar/0.9.1/rstar/primitives/struct.GeomWithData.html). Thanks for the PR @jblindsay, and apologies for first asking for, and then hijacking it. Hopefully the more general implementation addresses your need and allows for a little more flexibility.

PR #74, opened 2021-08-03
generic geom + data container

urschrei:
- [x ] I agree to follow the project's [code of conduct](https://github.com/georust/geo/blob/master/CODE_OF_CONDUCT.md).
---

This is a generic implementation of `PointWithData` and #73. <s>All fields are private and require getters, and</s> The geometry field can't be modified after creation in order to prevent a logic error if it's in an RTree. This PR also adds a deprecation notice to `PointWithData`.

For discussion:

- ~~The name~~
- ~~struct fields~~
- ~~Use of getters~~
- ~~What should we be deriving?~~
- ~~Can we derive serde traits?~~
- ~~Is there any need for a `Default` on either field?~~

### DIFF
--- a/rstar/src/primitives/geom_with_data.rs
+++ b/rstar/src/primitives/geom_with_data.rs
@@ -1,0 +1,136 @@
+use crate::envelope::Envelope;
+use crate::object::PointDistance;
+use crate::{object::RTreeObject, point::Point};
+
+/// An [RTreeObject] with a geometry and some associated data that can be inserted into an r-tree.
+///
+/// Often, adding metadata (like a database ID) to a geometry is required before adding it
+/// into an r-tree. This struct removes some of the boilerplate required to do so.
+///
+/// **Note:** while the container itself implements [RTreeObject], you will have to go through its
+/// [`geom`](GeomWithData::geom) method in order to access geometry-specific methods.
+///
+/// # Example
+/// ```
+/// use rstar::{RTree, PointDistance};
+/// use rstar::primitives::GeomWithData;
+///
+/// type RestaurantLocation = GeomWithData<[f64; 2], &'static str>;
+///
+/// let mut restaurants = RTree::new();
+/// restaurants.insert(RestaurantLocation::new([0.3, 0.2], "Pete's Pizza Place"));
+/// restaurants.insert(RestaurantLocation::new([-0.8, 0.0], "The Great Steak"));
+/// restaurants.insert(RestaurantLocation::new([0.2, -0.2], "Fishy Fortune"));
+///
+/// let my_location = [0.0, 0.0];
+///
+/// // Now find the closest restaurant!
+/// let place = restaurants.nearest_neighbor(&my_location).unwrap();
+/// println!("Let's go to {}", place.data);
+/// println!("It's really close, only {} miles", place.distance_2(&my_location));
+/// ```
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Default)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct GeomWithData<R: RTreeObject, T> {
+    geom: R,
+    /// Data to be associated with the geometry being stored in the [`RTree`](crate::RTree).
+    pub data: T,
+}
+
+impl<R: RTreeObject, T> RTreeObject for GeomWithData<R, T> {
+    type Envelope = R::Envelope;
+
+    fn envelope(&self) -> Self::Envelope {
+        self.geom.envelope()
+    }
+}
+
+impl<R: PointDistance, T> PointDistance for GeomWithData<R, T> {
+    fn distance_2(
+        &self,
+        point: &<Self::Envelope as Envelope>::Point,
+    ) -> <<Self::Envelope as Envelope>::Point as Point>::Scalar {
+        self.geom.distance_2(point)
+    }
+
+    fn contains_point(&self, p: &<Self::Envelope as Envelope>::Point) -> bool {
+        self.geom.contains_point(p)
+    }
+
+    fn distance_2_if_less_or_equal(
+        &self,
+        point: &<Self::Envelope as Envelope>::Point,
+        max_distance_2: <<Self::Envelope as Envelope>::Point as Point>::Scalar,
+    ) -> Option<<<Self::Envelope as Envelope>::Point as Point>::Scalar> {
+        self.geom.distance_2_if_less_or_equal(point, max_distance_2)
+    }
+}
+
+impl<R: RTreeObject, T> GeomWithData<R, T> {
+    /// Create a new [GeomWithData] struct using the provided geometry and data.
+    pub fn new(geom: R, data: T) -> Self {
+        Self { geom, data }
+    }
+
+    /// Get a reference to the container's geometry.
+    pub fn geom(&self) -> &R {
+        &self.geom
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::GeomWithData;
+    use crate::object::PointDistance;
+
+    use approx::*;
+
+    use crate::{primitives::Line, RTree};
+
+    #[test]
+    fn container_in_rtree() {
+        let line_1 = GeomWithData::new(Line::new([0.0, 0.0], [1.0, 1.0]), ());
+        let line_2 = GeomWithData::new(Line::new([0.0, 0.0], [-1.0, 1.0]), ());
+        let tree = RTree::bulk_load(vec![line_1, line_2]);
+
+        assert!(tree.contains(&line_1));
+    }
+
+    #[test]
+    fn container_edge_distance() {
+        let edge = GeomWithData::new(Line::new([0.5, 0.5], [0.5, 2.0]), 1usize);
+
+        assert_abs_diff_eq!(edge.distance_2(&[0.5, 0.5]), 0.0);
+        assert_abs_diff_eq!(edge.distance_2(&[0.0, 0.5]), 0.5 * 0.5);
+        assert_abs_diff_eq!(edge.distance_2(&[0.5, 1.0]), 0.0);
+        assert_abs_diff_eq!(edge.distance_2(&[0.0, 0.0]), 0.5);
+        assert_abs_diff_eq!(edge.distance_2(&[0.0, 1.0]), 0.5 * 0.5);
+        assert_abs_diff_eq!(edge.distance_2(&[1.0, 1.0]), 0.5 * 0.5);
+        assert_abs_diff_eq!(edge.distance_2(&[1.0, 3.0]), 0.5 * 0.5 + 1.0);
+    }
+
+    #[test]
+    fn container_length_2() {
+        let line = GeomWithData::new(Line::new([1, -1], [5, 5]), 1usize);
+
+        assert_eq!(line.geom().length_2(), 16 + 36);
+    }
+
+    #[test]
+    fn container_nearest_neighbour() {
+        let mut lines = RTree::new();
+        lines.insert(GeomWithData::new(
+            Line::new([0.0, 0.0], [1.0, 1.0]),
+            "Line A",
+        ));
+        lines.insert(GeomWithData::new(
+            Line::new([0.0, 0.0], [-1.0, 1.0]),
+            "Line B",
+        ));
+        let my_location = [0.0, 0.0];
+        // Now find the closest line
+        let place = lines.nearest_neighbor(&my_location).unwrap();
+
+        assert_eq!(place.data, "Line A");
+    }
+}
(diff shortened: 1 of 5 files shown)


urschrei:
@rmanoka @adamreichold @jblindsay please have a look and discuss.

urschrei:
One aspect of this I don't like is that because we impl `RTreeObject` on `GeomWithData` the "parent" struct has one set of methods for calculating distance / containment / etc, but you have to go through the setter for anything specific (e.g. `position`, `project_point`). It's not _terrible_, and we have a compiler to help out, but it's not the cleanest design.

rstar/src/primitives/geomwithdata.rs
```diff
@@ -0,0 +1,175 @@
...
+        self.0 = new_value;
+    }
+
+    /// Replace the container's data
+    pub fn replace_data(&mut self, new_value: T) -> T {
+        replace(&mut self.0, new_value)
+    }
+
+    /// Get a reference to the container's geometry
+    pub fn get_geom(&self) -> &P {
```
rmanoka, line 113: The rust convention is to just call this `geom()`.

rstar/src/primitives/geomwithdata.rs
```diff
@@ -0,0 +1,175 @@
...
+}
+
+impl<T, P> GeomWithData<T, P> {
+    /// Creates a new [GeomWithData] with the provided data.
+    pub fn new(data: T, geom: P) -> Self {
+        GeomWithData(data, geom)
+    }
+
+    /// Get a reference to the container's data
+    pub fn get_data(&self) -> &T {
```
rmanoka, line 98: Getters in rust are typically named without `get_` prefix.

rstar/src/primitives/geomwithdata.rs
```diff
@@ -0,0 +1,175 @@
...
+        &self.0
+    }
+
+    /// Set the container's data
+    pub fn set_data(&mut self, new_value: T) {
```
rmanoka, line 103: Given that we are trying to provide `get`, `set` and `replace`, we might as well make `data` a `pub` field?  Something like:

```rust
pub struct GeomWithData<T, R>{
  geom: T,
  pub data: R,
}
```

Feels a bit more ergonomic to access the data.

rstar/src/primitives/geomwithdata.rs
```diff
@@ -0,0 +1,175 @@
...
+/// restaurants.insert(RestaurantLocation::new("Pete's Pizza Place", [0.3, 0.2]));
+/// restaurants.insert(RestaurantLocation::new("The Great Steak", [-0.8, 0.0]));
+/// restaurants.insert(RestaurantLocation::new("Fishy Fortune", [0.2, -0.2]));
+///
+/// let my_location = [0.0, 0.0];
+///
+/// // Now find the closest restaurant!
+/// let place = restaurants.nearest_neighbor(&my_location).unwrap();
+/// println!("Let's go to {}", place.get_data());
+/// println!("It's really close, only {} miles", place.distance_2(&my_location));
```
rmanoka, line 31: Neat docs; nice example!

rmanoka:
Nice implementation @urschrei . I like that we're making the geom immutable (so only getter).

Regarding derives:  I think we should derive as many standard traits as possible as the users can anyway not implement them on this type.  So, Debug, Clone, Copy, PartialEq, Eq, Default.  Note that in all these cases, the derive macro only derives assuming both T, R also implement the trait which makes sense.  `PartialEq` is especially useful as it provides `RTree::contains`

`PartialOrd` and `Ord`:  I suppose most users would expect to have it sorted by the ordering of the geometry.  So, we should ensure that we declare the geom before the data and derive these so that the lexicographic ordering is intuitive.

`Hash` also seems like a standard trait that people like to derive.

adamreichold:
@rmanoka Considering the derives, I wonder we should actually not use the default derives but implement this manually to be restricted to the `geom` field and ignore the `data` field. This way `RTree::contains/remove` would work based solely on the geometry ignoring the associated data. I think there are use cases for both approaches.

rstar/src/primitives/geomwithdata.rs
```diff
@@ -0,0 +1,175 @@
+use crate::envelope::Envelope;
+use crate::object::PointDistance;
+use crate::{object::RTreeObject, point::Point};
+use std::mem::replace;
+
+/// An [RTreeObject] with some associated data and a geometry that can be inserted into an r-tree.
+///
+/// Often, adding metadata (like a database index) to a geometry is required before adding it
```
adamreichold, line 8: I would have expected a "database ID" here?

rstar/src/primitives/geomwithdata.rs
adamreichold: (I concur with the above, but even if we don't, I think it would be more idiomatic to provide something like `fn data_mut(&mut self) -> &mut T` instead of by value setter and replace method.)

rstar/src/primitives/geomwithdata.rs
urschrei: I stole them!

urschrei:
Thanks both, superficial comments are all addressed. @adamreichold your suggestion to restrict the derives to the geom field only makes sense to me. Can you and @rmanoka:
 
 1. sketch out which derives you want, and where, to make sure I haven't misunderstood you;
 2. make a decision about which option to go with

And then let's try to get this merged – if only because I feel bad about pushing James for a PR only to veer off in a (better, IMO) direction.

rstar/src/primitives/geomwithdata.rs
urschrei: My only reservation about making `data` public is that it's then public forever. In practice, it's unlikely that we'll be altering this struct in such a way as to restrict or otherwise redirect `data`, but a setter is more conservative. In any case, it's pub for now.

rstar/src/primitives/geomwithdata.rs
```diff
@@ -0,0 +1,163 @@
...
+    fn distance_2_if_less_or_equal(
+        &self,
+        point: &<Self::Envelope as Envelope>::Point,
+        max_distance_2: <<Self::Envelope as Envelope>::Point as Point>::Scalar,
+    ) -> Option<<<Self::Envelope as Envelope>::Point as Point>::Scalar> {
+        self.geom.distance_2_if_less_or_equal(point, max_distance_2)
+    }
+}
+
+impl<T, P> GeomWithData<T, P> {
```
adamreichold, line 94: Nittest of nits, but it might be confusing that the type parameter is called `P` here but `R` everywhere else. Also, you can construct the instance using `Self` as well, i.e.
```
pub fn new(data: T, geom: R) -> Self {
  Self { geom, data }
}
```

rstar/src/primitives/mod.rs
```diff
@@ -1,9 +1,11 @@
 //! Contains primitives ready for insertion into an r-tree.
 
+mod geomwithdata;
```
adamreichold, line 3: I think snake case `geom_with_data` as done for `point_with_data` would actually be more readable here.

rstar/src/primitives/geomwithdata.rs
urschrei: Oops, totally legit nit. Fixed!

rmanoka:
> @rmanoka Considering the derives, I wonder we should actually not use the default derives but implement this manually to be restricted to the `geom` field and ignore the `data` field. This way `RTree::contains/remove` would work based solely on the geometry ignoring the associated data. I think there are use cases for both approaches.

I suppose you are talking specifically about `PartialEq` and `Eq`?  I think the natural definition of these traits is to use both fields.  Further, if the user wants to check contains / remove with just the geometry, they could do it via the `nearest_neighbor` and `pop_nearest_neighbor` methods anyway.  Also, note that `PointWithData` currently follows this semantic, and [directly derives](https://github.com/georust/rstar/blob/14f019f2492fd5ca3f9003084fc3eb108e228eaf/rstar/src/primitives/point_with_data.rs#L27) all these traits

Stoeoef:
In regards to the `derive` discussion: I would also suggest to derive as much as possible and include both data and the geometry. That's *usually* what `PartialEq` is supposed to do, steering away from that would potentially be more confusing IMO.

There is also a simple workaround should you need a more specialized `contains` method by calling `rtree.locate_in_envelop(some_aabb).filter(some_special_filter)`.

Stoeoef (commented):
LGTM (besides the temporary test)

rstar/src/primitives/geom_with_data.rs
```diff
@@ -0,0 +1,164 @@
...
+
+        let my_location = [0.0, 0.0];
+
+        // Now find the closest line
+        let place = lines.nearest_neighbor(&my_location).unwrap();
+        println!("Let's go to {}", place.data);
+        println!(
+            "It's really close, only {} miles",
+            place.geom.distance_2(&my_location)
+        );
```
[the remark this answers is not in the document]
urschrei: Just needed to make sure it was working!

rstar/src/primitives/geom_with_data.rs
```diff
@@ -0,0 +1,164 @@
...
+        point: &<Self::Envelope as Envelope>::Point,
+        max_distance_2: <<Self::Envelope as Envelope>::Point as Point>::Scalar,
+    ) -> Option<<<Self::Envelope as Envelope>::Point as Point>::Scalar> {
+        self.geom.distance_2_if_less_or_equal(point, max_distance_2)
+    }
+}
+
+impl<T, R> GeomWithData<T, R> {
+    /// Creates a new [GeomWithData] with the provided data.
+    pub fn new(data: T, geom: R) -> Self {
```
michaelkirk, line 97:
```suggestion
    pub fn new(geom: R, data: T) -> Self {
```

Would this be a more intuitive param ordering?

Geom-With-Data takes a Geom, then a Data?

Not a big deal, so take it or leave it.

If you do this, you might also want to flip the generic param order too, so it's all consistent (`GeomWithData<R, T>`)

rstar/src/primitives/geomwithdata.rs
michaelkirk: I'll add one more voice...

```
fn data(&self) -> &T;
fn data_mut(&self) -> &mut T;
```

from: https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter
> With a few exceptions, the get_ prefix is not used for getters in Rust code.

Once you have a &mut there's not really a reason to have a set *and* a replace, plus then you can also do fun things like `swap` and `take` if you want them.

rstar/src/primitives/geom_with_data.rs
urschrei: Yep, this is a thoughtful improvement.

rstar/src/primitives/geomwithdata.rs
michaelkirk: oh sorry, I think I erroneously commented on an old diff here. I'm also ok with the `pub data: T,` solution you've adopted. 👍

urschrei:
- Temp test removed
- Order flipped so `geom` precedes `data`, as implied by the name
- Derives same as `PointWithData`, `serde` derives are gated. 
- Changelog updated
- Commits squashed

rmanoka (approved):
Looks great!  Thanks @urschrei

rstar/src/primitives/geom_with_data.rs
```diff
@@ -0,0 +1,136 @@
...
+/// let my_location = [0.0, 0.0];
+///
+/// // Now find the closest restaurant!
+/// let place = restaurants.nearest_neighbor(&my_location).unwrap();
+/// println!("Let's go to {}", place.data);
+/// println!("It's really close, only {} miles", place.distance_2(&my_location));
+/// ```
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Default)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct GeomWithData<R, T> {
```
michaelkirk, line 34:
```suggestion
pub struct GeomWithData<R: RTreeObject, T> {
```

One more minor thing - folks might get a *slightly* clearer error message if they mix up these params if we also constrain the struct definition, rather than only constraining the impls.

I can't think of a reason we'd want someone to be able to build a GeomWithData where R was *not* an RTreeObject, right?

urschrei: Something like 335a029? I _did_ wonder why that hadn't been done for `PointWithData`…

michaelkirk: Yep, that's exactly what I had in mind!

michaelkirk (approved):

Stoeoef (approved):

urschrei:
Bors r+

PR #83, opened 2022-01-14
enable compilation for no_std

wucke13:
---

This is basically a redo of https://github.com/georust/rstar/pull/27 . ~~I'm not sure whether we should have the `alloc` feature or not. In the best case, we would haven an rstar implementation which can work without alloc - but my understanding of this implementation is rather limited. Currently, one needs to either pick the `std` or the alloc feature, picking none of them results in compilation failure. Maybe we can resort to `heapless::Vec` in the future, but this for sure requires further modifications.~~ This does not include any new features. Compilation is supported for both `std` and `no_std`, but the latter requires `alloc`. For validation, I did this:

+ `cargo build --target thumbv7em-none-eabihf`
+ `cargo build`
+ `cargo build --release`
+ `cargo test`
+ `cargo test --release`

### DIFF
--- a/rstar/Cargo.toml
+++ b/rstar/Cargo.toml
@@ -16,8 +16,8 @@ maintenance = { status = "actively-developed" }
 
 [dependencies]
 heapless = "0.6"
-num-traits = "0.2"
-serde = { version = "1.0", optional = true, features = ["derive"] }
+num-traits = { version = "0.2", default-features = false, features = ["libm"] }
+serde = { version = "1.0", optional = true, default-features = false, features = ["alloc", "derive"] }
 smallvec = "1.6"
 
 [features]
--- a/rstar/src/algorithm/bulk_load/bulk_load_sequential.rs
+++ b/rstar/src/algorithm/bulk_load/bulk_load_sequential.rs
@@ -4,6 +4,11 @@ use crate::object::RTreeObject;
 use crate::params::RTreeParams;
 use crate::point::Point;
 
+use alloc::{vec, vec::Vec};
+
+#[allow(unused_imports)] // Import is required when building without std
+use num_traits::Float;
+
 use super::cluster_group_iterator::{calculate_number_of_clusters_on_axis, ClusterGroupIterator};
 
 fn bulk_load_recursive<T, Params>(elements: Vec<T>, depth: usize) -> ParentNode<T>
@@ -47,7 +52,7 @@ struct PartitioningTask<T: RTreeObject, Params: RTreeParams> {
     work_queue: Vec<PartitioningState<T>>,
     depth: usize,
     number_of_clusters_on_axis: usize,
-    _params: std::marker::PhantomData<Params>,
+    _params: core::marker::PhantomData<Params>,
 }
 
 impl<T: RTreeObject, Params: RTreeParams> Iterator for PartitioningTask<T, Params> {
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,6 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     needs:
       - rstar
+      - no_std
     steps:
       - name: Mark the job as a success
         if: success()
@@ -40,3 +41,24 @@ jobs:
       - run: cargo build-all-features
       - run: cargo test-all-features
       - run: cargo build -p rstar-benches
+
+  no_std:
+    name: rstar no_std test
+    runs-on: ubuntu-latest
+    env: 
+      NO_STD_TARGET: aarch64-unknown-none
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+      - name: Install stable toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          target: ${{env.NO_STD_TARGET}}
+          toolchain: stable
+          override: true
+      - name: Run cargo build for ${{env.NO_STD_TARGET}}
+        uses: actions-rs/cargo@v1
+        with:
+          command: build
+          args: --package rstar --target ${{env.NO_STD_TARGET}}
--- a/rstar/CHANGELOG.md
+++ b/rstar/CHANGELOG.md
@@ -2,6 +2,8 @@
 ## Changed
 - Removed dependency on `pdqselect` ([PR](https://github.com/georust/rstar/pull/85))
 - New **minimal supported rust version (MSRV): 1.49.0**
+- Replace all usages of `std` with `core` & `alloc` to make `rstar` fit for
+  `no_std`. ([PR](https://github.com/georust/rstar/pull/83))
 
 # 0.9.2
 - Add `RTree::drain_*` methods to remove and drain selected items. ([PR](https://github.com/georust/rstar/pull/77))
--- a/rstar/README.md
+++ b/rstar/README.md
@@ -20,6 +20,7 @@ A flexible, n-dimensional [r*-tree](https://en.wikipedia.org/wiki/R*_tree) imple
    - Rectangles
  - Small number of dependencies
  - Serde support with the `serde` feature
+ - `no_std` compatible (but requires [`alloc`](https://doc.rust-lang.org/alloc/))
 
 # Benchmarks
 All benchmarks are performed on a i7-8550U CPU @ 1.80Ghz and with uniformly distributed points. The underlying point type is `[f64; 2]`.
--- a/rstar/src/aabb.rs
+++ b/rstar/src/aabb.rs
@@ -156,7 +156,7 @@ where
             max = max * max;
             min = min * min;
             if max < min {
-                std::mem::swap(&mut min, &mut max);
+                core::mem::swap(&mut min, &mut max);
             }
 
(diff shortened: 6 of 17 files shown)


rstar/Cargo.toml
```diff
@@ -16,14 +16,16 @@ maintenance = { status = "actively-developed" }
...
-serde = { version = "1.0", optional = true, features = ["derive"] }
+serde = { version = "1.0", optional = true, default-features = false, features = ["derive"] }
 smallvec = "1.6"
 
 [features]
-default = []
+default = ["std"]
 debug = []
+std = [ "num-traits/default" ]
+alloc = [ "num-traits/libm" ]
```
Stoeoef, line 28: Is there a specific need for the `alloc` feature? Users might forget to include it after choosing `--no-default-features`.

We could, instead, replace any `cfg(feature = alloc)` with `cfg(not(feature = "std"))` and _always_ use `num_traits` with the `libm` feature.

However, I'm okay with leaving this as is if there's some reason for it that I'm not aware of. Working with no-std is fairly new to me, please excuse if I'm not aware of conventions and best practices in this area.

rstar/src/test_utilities.rs
```diff
@@ -4,6 +4,8 @@ use rand::distributions::Uniform;
 use rand::{Rng, SeedableRng};
 use rand_hc::Hc128Rng;
 
+use crate::lib::*;
```
Stoeoef, line 7: I believe this import shouldn't be required as `test_utilities` is only used with the `test` feature.

Stoeoef (commented):
Thanks for looking into this!

Changes look good to me besides two small comments.

Remaining TODOs would, in my opinion, include:
- [ ] Add documentation in `lib.rs` on how to enable `no-std` support
- [ ] Add documentation in `README.md` on how to enable `no-std` support
- [ ] Add a build command to the CI environment. I think `cargo build --package rstar --no-default-features` (possibly with an additional `features=alloc`) should be executed with every CI run.

rstar/Cargo.toml
wucke13: I believe omitting the `alloc` should be possible, exactly the way you proposed. The question here is: can the `rstar` crate deliver any value as library if all things which depend on `alloc` are removed? If not, I'm happy to factor out the `alloc` feature just as you described.

rstar/src/test_utilities.rs
wucke13: Good catch!

rstar/Cargo.toml
Stoeoef: From my understanding, not having `alloc` would require to "reserve" a fixed amount of memory on the stack and using that for representing the tree.
This would require to represent the tree as part of a continuous segment of memory - it's currently modelled as a "real" tree on the heap (See struct `ParentNodeData`).
While that is something that could be interesting (e.g. just for the performance benefits it may bring), it's also a lot of work and restructuring that will be required. Thus, I wouldn't bet on this feature (unless we find someone wanting to work on this) and would be in **favour of leaving out `alloc` as a feature**.

Sorry for the lengthy explanation - I'm giving it in the hope to be corrected in case anything is represented inaccurately.

wucke13: Ok, then there is some things to discuss left: AFAIK there is no negative feature lookup, e. g. when there is no specific feature for `no_std` support, we can not conditionally depend on `alloc` in the Cargo.toml. However, we can also get away without using features at all! We can just always depend on `alloc` and `core`, and the lib is unconditionally always `no_std`. This should not affect downstream consumers, everybody who either

+ offers an `alloc` implementation
+ compiles with std

can use the rstar. Are you fine if we pursue this way?

Stoeoef: Good observation! I didn't look at it from this angle.
Let's go for the feature-less approach that you have described.

wucke13: Ok, I updated the commit accordingly. Please revisit your TODO items, they probably changed now?

wucke13:
@Stoeoef regarding checks of the `no_std` compatibility, maybe we should use [`cargo-nono`](https://github.com/hobofan/cargo-nono), it seems to be made precisely for this use case.

Stoeoef:
Thanks for the updates!

However, I would still think that a mention in `lib.rs` and `README.md` (the one in `rstar`) would be appropriate. Some short sentence like

```markdown
# no-std support
rstar supports `no-std` environments by default but requires the `alloc` crate.
```

wucke13:
I added a notice to the list of features (at least to me, `no_std` compatibiliy is a feaure :). A further note: When I run `cargo fmt`, even lines which I did not touch are changed. Maybe `cargo fmt --check`would also make sense in the CI?

Stoeoef:
> @Stoeoef regarding checks of the `no_std` compatibility, maybe we should use [`cargo-nono`](https://github.com/hobofan/cargo-nono), it seems to be made precisely for this use case.

I've checked `cargo-nono` . Unfortunately, it doesn't produce reliable results for `rstar` - it marks some of its dependencies as `no-std` even though they are fine (e.g. smallvec).

Thus, I think an additional check in the CI configuration would be more appropriate. That should consistently catch any conflicting no-std dependency.

Please let me know if I can help you out with anything CI related and I'll see what I can do!

Stoeoef:
> I added a notice to the list of features (at least to me, `no_std` compatibiliy is a feaure :). A further note: When I run `cargo fmt`, even lines which I did not touch are changed. Maybe `cargo fmt --check`would also make sense in the CI?

Thanks for adding the entry to the changelog, I agree that this should be included as a feature.
In regards to `cargo fmt`: Captain hindsight agrees with you. However, let's keep this PR focussed for now. Feel free to revert / keep any formatting changes due to `cargo fmt` as you like.

rstar/src/algorithm/bulk_load/bulk_load_sequential.rs
```diff
@@ -4,6 +4,11 @@ use crate::object::RTreeObject;
 use crate::params::RTreeParams;
 use crate::point::Point;
 
+use alloc::{vec, vec::Vec};
+
+#[cfg(not(test))]
+use num_traits::Float;
```
Stoeoef, line 10: I think this is outdated now - I think every occurrence of `#[cfg(not(test))]`  (there should be three of them) can be omitted.

wucke13: If I remove it and run `cargo test` from the workspace root, I get

```console
warning: unused import: `num_traits::Float`
 --> rstar/src/algorithm/bulk_load/bulk_load_sequential.rs:8:5
  |
8 | use num_traits::Float;
  |     ^^^^^^^^^^^^^^^^^
  |
  = note: `#[warn(unused_imports)]` on by default

warning: unused import: `num_traits::Float`
 --> rstar/src/algorithm/bulk_load/cluster_group_iterator.rs:4:5
  |
4 | use num_traits::Float;
  |     ^^^^^^^^^^^^^^^^^

warning: unused import: `num_traits::Float`
  --> rstar/src/algorithm/removal.rs:10:5
   |
10 | use num_traits::Float;
   |     ^^^^^^^^^^^^^^^^^

warning: `rstar` (lib) generated 3 warnings
```

Are you sure?

wucke13:
> Please let me know if I can help you out with anything CI related and I'll see what I can do!

Please have a look at the workflow file, I believe that this could work but I'm not sure about my usage of `env`

rstar/src/algorithm/bulk_load/bulk_load_sequential.rs
Stoeoef: I was a little unclear there - the _import_ is still required, the `cfg(not(test))` doesn't seem to make any difference any more. I'm getting warnings with and without the `cfg(not...)` block when running `cargo test`.

Unless you have a different idea I would suggest to use this instead:

```rust
#[allow(unused_imports)] // Import is required when building without std
use num_traits::Float;
```

wucke13: Ah, you where probably running `cargo test` from the workspace root? When you run `cargo test` from the workspace root, I get the warning. When I run `cargo test` from the `rstar` folder, I do not get the warning. Either way, yes, I'll go for your fix.

wucke13: Done

wucke13:
Lets try that again :)

wucke13:
bors try

Stoeoef:
I'm a little concerned that bors will think everything has passed, even though the std step failed. I'll investigate this while retrying

bors try

wucke13:
That was my error, I forgot to add the new job to the `needs` section of bors despite the angry ALL CAPSLOCK WARNING WHICH TOLD ME TO DO SO.

Stoeoef:
bors try

wucke13:
I think we have it. I made another small adjustment in the naming of the ci job. If you have nothing left to do, I guess we are good to go?

Stoeoef:
Yes!  I'm happy as a clam.

Thanks for your work on this PR!

bors r+

PR #85, opened 2022-01-21
Remove pdqselect dependency

Stoeoef:
---

Pros:
- Fewer dependencies!
- Switching to `select_unstable` (rather than `stable`) should be faster. I can run a few benchmarks to back this up with numbers if desired. A stable sort is not required for the algorithm to work correctly

Cons:
- Bumps the minimal supported rust version (MSRV)  to 1.49

------------------
What's the thought about the MSRV? Is it okay to rely on 1.49? V1.49 was released on 2020-12-31


Stoeoef:
Thanks! 
bors r+

PR #86, opened 2022-03-12
Geo points and custom distance metric

DXist:
---
This PR is a demo of an attempt to use rstar to store geo points (with longitude and latitude as x, y values) and run nearest neighbour algorithm between close points.

To implement the demo in external crate I have exposed PointExp trait to public interface and disabled default PointExp implementation for Point types except arrays and tuples.

I added the demo to `rstar/src/geopoint.rs`.

Custom `sub` method calculates approximate distance between two points on WGS84 ellipsoid model of the Earth(https://en.wikipedia.org/wiki/Earth_radius#Meridional), using the code from https://github.com/vipera/cheap-ruler-rs . This allows to calculate distances with acceptable precision up to ~500km which is good enough for my use case.
The method returns deltas along longitudes and latitudes scaled to distances in meters.

As I understand `PointExt` should be private. Is it better to introduce a public trait for coordinate metric distances and implement `PointExt::sub` method via something like `CoordMetricDistance::delta_between_points`?

### DIFF
--- a/rstar/src/geopoint.rs
+++ b/rstar/src/geopoint.rs
@@ -1,0 +1,112 @@
+use crate::PointExt;
+use geo_types;
+
+type LonLatFloat = f32;
+
+#[derive(PartialEq, Clone, Copy, Debug, Default)]
+struct GeoPoint(geo_types::Point<LonLatFloat>);
+
+impl GeoPoint {
+    const RE: f64 = 6378.137; // equatorial radius in km
+    const FE: f64 = 1.0 / 298.257223563; // flattening
+    const E2: f64 = Self::FE * (2.0 - Self::FE);
+
+    pub fn new(x: LonLatFloat, y: LonLatFloat) -> Self {
+        GeoPoint(geo_types::Point::new(x, y))
+    }
+
+    fn x(&self) -> LonLatFloat {
+        self.0.x()
+    }
+
+    fn y(&self) -> LonLatFloat {
+        self.0.y()
+    }
+
+    fn long_diff(a: LonLatFloat, b: LonLatFloat) -> LonLatFloat {
+        let threesixty = 360.;
+        let diff = a - b;
+        diff - ((diff / threesixty).round() * threesixty)
+    }
+
+    fn multipliers_for_lonlat_converstion(latitude: LonLatFloat) -> (LonLatFloat, LonLatFloat) {
+        let one = 1.;
+        let e2 = Self::E2 as LonLatFloat;
+
+        // Curvature formulas from https://en.wikipedia.org/wiki/Earth_radius#Meridional
+        let coslat = latitude.to_radians().cos();
+        let w2 = one / (one - e2 * (one - coslat * coslat));
+        let w = w2.sqrt();
+
+        // multipliers for converting longitude and latitude degrees into distance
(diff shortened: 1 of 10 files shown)


michaelkirk:
Hello! Were you aware of the optional rstar integration already built into geo-types? Or is this trying to achieve something else?

https://github.com/georust/geo/blob/master/geo-types/src/point.rs#L552

DXist:
Hi @michaelkirk , I want to get nearest neighbours and their geo distances in meters for relatively close geo points. I am going to use these distances to obtain lower and upper bounds in [Incremental Euclidean Restriction ](https://www.vldb.org/conf/2003/papers/S24P02.pdf) algorithm and then find nearest neighbours on road network.

rstar/src/geopoint.rs
```diff
@@ -0,0 +1,112 @@
...
+        let re = Self::RE as LonLatFloat;
+        let mul = (1000. as LonLatFloat).to_radians() * re;
+        let kx = mul * dkx;
+        let ky = mul * dky;
+        (kx, ky)
+    }
+}
+
+impl PointExt for GeoPoint {
+    fn sub(&self, other: &Self) -> Self {
```
adamreichold, line 58: Isn't this definition inconsistent with the various default implementations defined by the `PointExt` trait?

adamreichold:
> specifically, the approach I'd personally take is, to have my application [proj](https://github.com/georust/proj)ect my points to a cartesian space (maybe using proj), and then trust the rtree to work with boring normal math.

I am not sure about the performance implications, but I think this might be preferable. Using [`PointWithData`](https://docs.rs/rstar/latest/rstar/primitives/struct.PointWithData.html), the original coordinates could be stored alongside the projected ones so that there are no roundtrip inaccuracies for the results of the neighbourhood query.

DXist:
> So you are dealing with lat/lon points, but we want to find neighbors based on meters.
> 
> I'm not familiar enough to know if this is reasonable. I've only ever used an R-Tree with boring cartesian math. So I'll have to wait for someone more knowledgable to weigh in.
> 
> edit: specifically, the approach I'd personally take is, to have my application project my points to a cartesian space (maybe using [proj](https://github.com/georust/proj)), and then trust the rtree to work with boring normal math.

Projections like Universal Mercator Projection work good for local data on city or small region level. This approach becomes inconvenient when geo coordinates are expected to be global. For UTM there are 60 zones in 2 hemispheres that require 120 local Rtrees to store global data. Zone partitioning introduces discontinuity and complicates processing geo data around the boundaries.

rstar/src/geopoint.rs
DXist: It's just a way how I introduced custom distance implementation for lon/lat coordinates.

michaelkirk: Thank you for the link to the paper! I'll check it out.

DXist:
With precomputed coefficients NN queries with custom distance method are about ~18 times slower than original implementation.

DXist:
I've decided to convert coordinates to 3D ECEF coordinate system and use euclidean distances.
My benchmark shows that it's ~3 times faster to construct and ~21 times faster to query nearest neighbours than using custom distance metric with 2D lon/lat coordinates.

Thank you for your comments! I'm closing this issue.

PR #87, opened 2022-03-20
Upgrade to heapless v0.7.10 and update MSRV to 1.51.0

rye:
---

[Rust 1.51](https://blog.rust-lang.org/2021/03/25/Rust-1.51.0.html) has been released for nearly 12 months now.

As the community upgrades through and past this Rust version, some required dependencies of `rstar` (namely, `heapless` v0.6, which itself required the `as-slice` and `generic-array` dependencies before Rust 1.51) have since released updates to make use of the new Const Generics MVP functionality that was released with Rust 1.51.

To remove a few outdated dependencies and upgrade the codebase to the latest version of `heapless`, this PR does the following:

- Upgrades the `heapless` dependency from 0.6 to 0.7.

  This broke my local compile, as `heapless`' signature for the `heapless::binary_heap::BinaryHeap` type changed, and the `heapless::consts::U32` can now be represented with a plain old `32_usize`. I made changes to `src/algorithm/nearest_neighbor.rs` to fix the compile, and this resolved the compile failure. The tests now seem to pass.
  
- Updates the CI workflows file to test versions as far back as Rust 1.51 (but not Rust 1.50), as well as two more recent MSRVs.

- Updates the `rstar/Cargo.toml` file to specify a `rust-version` key to indicate that Rust 1.51 is the Minimum Supported Rust Version (MSRV). This will display a warning on versions of cargo older than that which was distributed with Rust 1.56, as the `rust-version` key was added in that release.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -32,8 +32,9 @@ jobs:
     strategy:
       matrix:
         container_image:
-          - "georust/geo-ci:rust-1.50"
           - "georust/geo-ci:rust-1.51"
+          - "georust/geo-ci:rust-1.58"
+          - "georust/geo-ci:rust-1.59"
     container:
       image: ${{ matrix.container_image }}
     steps:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -31,7 +31,9 @@ jobs:
         working-directory: rstar
     strategy:
       matrix:
-        container_image: ["georust/geo-ci:rust-1.50", "georust/geo-ci:rust-1.51"]
+        container_image:
+          - "georust/geo-ci:rust-1.50"
+          - "georust/geo-ci:rust-1.51"
     container:
       image: ${{ matrix.container_image }}
     steps:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -32,7 +32,9 @@ jobs:
     strategy:
       matrix:
         container_image:
+          # Minimum supported rust version (MSRV)
           - "georust/geo-ci:rust-1.51"
+          # Two most recent releases - older ones are omitted.
           - "georust/geo-ci:rust-1.58"
           - "georust/geo-ci:rust-1.59"
     container:
--- a/rstar/Cargo.toml
+++ b/rstar/Cargo.toml
@@ -8,6 +8,7 @@ repository = "https://github.com/georust/rstar"
 license = "MIT/Apache-2.0"
 readme = "README.md"
 edition = "2018"
+rust-version = "1.51"
 keywords = ["rtree", "r-tree", "rstar", "spatial", "nearest-neighbor"]
 categories = ["data-structures", "algorithms"]
 
--- a/rstar/Cargo.toml
+++ b/rstar/Cargo.toml
@@ -15,7 +15,7 @@ categories = ["data-structures", "algorithms"]
 maintenance = { status = "actively-developed" }
 
 [dependencies]
-heapless = "0.6"
+heapless = "0.7"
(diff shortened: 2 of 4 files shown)


rmanoka:
Just a minor nit: pls  annotate the chanelog entry, mentioning this as a breaking change.  Will bump the version appropriately when publishing the next release.

Looks good to me otherwise.

rstar/Cargo.toml
```diff
@@ -8,14 +8,15 @@ repository = "https://github.com/georust/rstar"
...
+rust-version = "1.51"
 keywords = ["rtree", "r-tree", "rstar", "spatial", "nearest-neighbor"]
 categories = ["data-structures", "algorithms"]
 
 [badges]
 maintenance = { status = "actively-developed" }
 
 [dependencies]
-heapless = "0.6"
+heapless = "0.7"
```
michaelkirk, line 19: Could you specify the full version you've tested against here?

see https://users.rust-lang.org/t/psa-please-specify-precise-dependency-versions-in-cargo-toml/71277 for rationale.

It's something I've only recently realized that we've gotten wrong historically, but I'd like to get on track with it.

.github/workflows/test.yml
```diff
@@ -31,7 +31,12 @@ jobs:
         working-directory: rstar
     strategy:
       matrix:
-        container_image: ["georust/geo-ci:rust-1.50", "georust/geo-ci:rust-1.51"]
+        container_image:
+          # Minimum supported rust version (MSRV)
```
michaelkirk, line 35: 👍

michaelkirk (changes requested):

michaelkirk:
> Just a minor nit: pls annotate the chanelog entry, mentioning this as a breaking change. 

Historically, there's been a lot of discussion on wether MSRV bump should be considered a breaking change. e.g. https://users.rust-lang.org/t/rust-version-requirement-change-as-semver-breaking-or-not/20980/39

I don't really understand how the rust-version constraint added to Cargo.toml since 1.56 will affect this, but since we're still supporting toolchains older than that, I suspect it's yet irrelevant.

rstar/Cargo.toml
rye: Sure! I was initially thinking of including the full version but when I saw it was just specified to two components I leaned towards keeping it consistent.

Since I tested against v0.7.10, I put that here.

rmanoka:
I'm sorry if I wasn't clear.  Upgrading heapless is what I considered breaking, not the msrv change.

rmanoka:
I see I may be mistaken here.  Is the heapless change visible in public API?  If not it's not really breaking change, and we don't need to mention in the changelog.

Regarding, msrv, yeah it's debatable whether to call it a breaking change.  @rye sorry for the confusion, but could you revert back to your previous changelog message?

rye:
@rmanoka no problem, I have reverted back to the original state.

To be clear:

- The change to `heapless` is not visible in the public API.

- The only place the `heapless` crate is used is in the `algorithm::nearest_neighbor` module, as an inner type in the `SmallHeap::Stack` type. This enum (`SmallHeap`) is not `pub` — it seems to be an internal type only. (You can try and search for `BinaryHeap` or `SmallHeap` or `Stack` on [docs.rs](https://docs.rs/rstar/latest/rstar/) to confirm this.)

- A separate `BinaryHeap`, the one from `alloc::collections::binary_heap` _is_ used and visible in the public API in the same module, but that's not the one that's changed here.

michaelkirk:
Thanks @rye!

bors r+

rmanoka:
may be try more politely?  please:
bors r+

PR #93, opened 2022-04-30
Fix panic when setting RTreeParams MIN_SIZE to 1

### LINKED ISSUE #92: "capacity overflow" on remove when RTreeParams::MIN_SIZE = 1
It appears that removing a object when the MIN_SIZE = 1 causes `capacity overflow` panic.

Could anyone elaobrate why? mabye this behaviour should be documented?

The code below reproduces the issue for me, on version 0.9.3

```rust
use rstar::{RStarInsertionStrategy, RTree, RTreeParams};

pub struct WierdParams;

impl RTreeParams for WierdParams {
    const MIN_SIZE: usize = 1;
    const MAX_SIZE: usize = 4;
    const REINSERTION_COUNT: usize = 1;
    type DefaultInsertionStrategy = RStarInsertionStrategy;
}

fn main() {
    let mut items: Vec<[f32; 2]> = Vec::new();
    for i in 0..2 {
        items.push([i as f32, i as f32]);
    }
    let mut tree: RTree<_, WierdParams> = RTree::bulk_load_with_params(items);
    println!("{:?}", tree.remove(&[1.0, 1.0]).is_some());
}
```

urschrei:
rstar's capacity-calculation logic would attempt to set the underlying vec's capacity to 2^64 -1 as a result, causing a panic because this exceeds the max value of isize on 64-bit platforms. This fix adds a check for MIN_SIZE = 1 when calculating capacity.

The fix currently sets the max depth to 1 – should it be higher?

Fixes #92

---

### DIFF
--- a/rstar/src/algorithm/removal.rs
+++ b/rstar/src/algorithm/removal.rs
@@ -58,7 +58,7 @@ where
         let original_size = replace(rtree.size_mut(), 0);
 
         let m = Params::MIN_SIZE;
-        let max_depth = (original_size as f32).log(m as f32).ceil() as usize;
+        let max_depth = (original_size as f32).log(m.max(2) as f32).ceil() as usize;
         let mut node_stack = Vec::with_capacity(max_depth);
         node_stack.push((root, 0, 0));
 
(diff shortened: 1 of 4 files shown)


urschrei:
Blocked on https://github.com/georust/rstar/issues/94

rstar/src/algorithm/removal.rs
```diff
@@ -58,7 +58,11 @@ where
         let original_size = replace(rtree.size_mut(), 0);
 
         let m = Params::MIN_SIZE;
-        let max_depth = (original_size as f32).log(m as f32).ceil() as usize;
```
rmanoka, line 61: We could replace the `m as f32` with `m.max(2) as f32` and it should fix the issue.  The value is used as an estimate for the depth of the tree as we don't store it anywhere.

rmanoka (approved):
Only minor nits!  Thanks for fixing @urschrei

urschrei:
bors r=rmanoka

PR #95, opened 2022-05-02
clarify MIN_SIZE constraints in docs

michaelkirk:
- [-] I added an entry to `rstar/CHANGELOG.md` if knowledge of this change could be valuable to users.
---

Addendum to #93 

Are these the docs that lead you astray in #92 @laundmo?

### DIFF
--- a/rstar/src/params.rs
+++ b/rstar/src/params.rs
@@ -35,10 +35,12 @@ use crate::{Envelope, Point, RTree, RTreeObject};
 /// # }
 /// ```
 pub trait RTreeParams: Send + Sync {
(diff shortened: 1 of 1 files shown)


laundmo:
If i understand #93 correctly the panic with a min-size of 1 is fixed. why does this require a docs change? the docs were clear before, just lacking the info that you couldn't use min-size 1.

urschrei:
Because I added an additional check asserting that MIN_SIZE is greater than zero. This was implicit before, but with this change is both explicit **and** documented.

rmanoka (approved):

michaelkirk:
bors r=rmanoka

PR #110, opened 2023-02-01
perf: minimize iterations over children in `IntersectionIterator`

kylebarron:
---

I was reading through the code and noticed that this was a double for loop, but `children2` doesn't depend on `child1` at all, so it doesn't need to be in the for loop. You do have to collect the items into a vector because you can't use the iterator more than once, but seems better than recomputing it every loop iteration?

### DIFF
--- a/rstar/src/algorithm/intersection_iterator.rs
+++ b/rstar/src/algorithm/intersection_iterator.rs
@@ -5,6 +5,7 @@ use crate::RTreeNode::*;
 use crate::RTreeObject;
 
 use alloc::vec::Vec;
+use core::mem::take;
 
 #[cfg(doc)]
 use crate::RTree;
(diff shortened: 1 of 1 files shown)


rstar/src/algorithm/intersection_iterator.rs
```diff
@@ -50,13 +50,14 @@ where
...
-        for child1 in children1 {
-            let children2 = parent2
-                .children()
-                .iter()
-                .filter(|c2| c2.envelope().intersects(&parent1.envelope()));
+        let children2: Vec<&RTreeNode<U>> = parent2
+            .children()
+            .iter()
+            .filter(|c2| c2.envelope().intersects(&parent1.envelope()))
+            .collect();
```
adamreichold, line 57: I think it would be preferable to store the `Vec` inside the iterator itself and reuse the allocation instead of collecting a fresh one. (Depending on whether this is possible or not, another option might be to use a `SmallVec` to avoid heap allocations in some cases.)

adamreichold (approved):
@kylebarron I took the liberty to rebase this and add an additional commit to resolve https://github.com/georust/rstar/pull/110#discussion_r1092818788.

@urschrei Could you have a look at the result? I would be happy to merge this as is.

urschrei:
Lgtm!

urschrei:
Bors r=adamreichold

PR #117, opened 2023-04-07
Use eager leaf envelope computation

urschrei:
---

### DIFF
--- a/rstar/src/rtree.rs
+++ b/rstar/src/rtree.rs
@@ -208,7 +208,7 @@ where
     /// # Runtime
     /// Bulk loading runs in `O(n * log(n))`, where `n` is the number of loaded
     /// elements.
-    pub fn bulk_load(elements: Vec<T>) -> Self {
+    pub fn bulk_load<I: IntoIterator<Item = T>>(elements: I) -> Self {
         Self::bulk_load_with_params(elements)
     }
 }
--- a/rstar/src/rtree.rs
+++ b/rstar/src/rtree.rs
@@ -235,7 +235,7 @@ where
     ///
     /// For more information refer to [RTree::bulk_load]
     /// and [RTreeParams].
-    pub fn bulk_load_with_params(elements: Vec<T>) -> Self {
+    pub fn bulk_load_with_params<I: IntoIterator<Item = T>>(elements: I) -> Self {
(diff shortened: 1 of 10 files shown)


rstar/src/rtree.rs
```diff
@@ -235,7 +235,7 @@ where
     ///
     /// For more information refer to [RTree::bulk_load]
     /// and [RTreeParams].
-    pub fn bulk_load_with_params(elements: Vec<T>) -> Self {
+    pub fn bulk_load_with_params<I: IntoIterator<Item = T>>(elements: I) -> Self {
```
adamreichold, line 238: `RTree::bulk_load` should also have `Vec<T>` replaced by `I: IntoIterator<Item = T>` to avoid needlessly allocating a `Vec` which will be deconstructed anyway.

urschrei: Done, although it's now causing a test to stop compiling due to a tricky type inference failure. See 2dce546

urschrei:
Test failures are spurious: due my disabling a test which can't run due to a type inference issue.

rstar/src/rtree.rs
adamreichold: Indeed, the same `.iter().cloned()` change which helps the benchmark will also fix the test though

```diff
diff --git a/rstar/src/algorithm/bulk_load/bulk_load_sequential.rs b/rstar/src/algorithm/bulk_load/bulk_load_sequential.rs
index a2ff07c..eae997a 100644
--- a/rstar/src/algorithm/bulk_load/bulk_load_sequential.rs
+++ b/rstar/src/algorithm/bulk_load/bulk_load_sequential.rs
@@ -154,7 +154,7 @@ mod test {
         P: RTreeObject + Send + Sync + Eq + Clone + Debug + Hash + 'static,
         P::Envelope: Send + Sync,
     {
-        let tree = RTree::bulk_load(points.into());
+        let tree = RTree::bulk_load(points.iter().cloned());
         let set1: HashSet<_> = tree.iter().collect();
         let set2: HashSet<_> = points.iter().collect();
         assert_eq!(set1, set2);
```

(The `.into()` call was basically `.to_vec()` which would also work, but allocate the unnecessary temporary `Vec`.)

urschrei: Great, fixed.

adamreichold:
Closing since we went for #118 instead.

PR #118, opened 2023-04-16
Add GeomWithCachedEnvelope combinator to support memoizing envelope computations.

### LINKED ISSUE #108: Document in `RTreeObject` that `envelope()` should be cached
I've been testing inserting geometric objects into rstar. I noticed a ~20x speedup with some test data (300k `MultiPolygon`s) between inserting _geometries_ where `RTreeObject::envelope()` was implemented with a direct call to `bounding_rect` to inserting _wrapper structs_ that include both the geometry and a pre-cached envelope. This 20x speedup includes the time in the latter case to loop over the geometries and create each's envelope.

In retrospect, it's not too surprising that `envelope()` would need to be called multiple times, but I didn't see a mention of it on the [`RTreeObject`](https://docs.rs/rstar/latest/rstar/trait.RTreeObject.html) page.

adamreichold:

Closes #108

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -33,10 +33,10 @@ jobs:
       matrix:
         container_image:
           # Minimum supported rust version (MSRV)
-          - "georust/geo-ci:rust-1.51"
-          # Two most recent releases - older ones are omitted.
-          - "georust/geo-ci:rust-1.58"
-          - "georust/geo-ci:rust-1.59"
+          - "georust/geo-ci:rust-1.63"
+          # Two most recent releases - we omit older ones for expedient CI
+          - "georust/geo-ci:rust-1.65"
+          - "georust/geo-ci:rust-1.66"
     container:
       image: ${{ matrix.container_image }}
     steps:
@@ -50,7 +50,7 @@ jobs:
   no_std:
     name: rstar no_std test
     runs-on: ubuntu-latest
-    env: 
+    env:
       NO_STD_TARGET: aarch64-unknown-none
     steps:
       - name: Checkout repository
--- a/rstar-benches/benches/benchmarks.rs
+++ b/rstar-benches/benches/benchmarks.rs
@@ -27,7 +27,7 @@ impl RTreeParams for Params {
 const DEFAULT_BENCHMARK_TREE_SIZE: usize = 2000;
 
 fn bulk_load_baseline(c: &mut Criterion) {
-    c.bench_function("Bulk load baseline", move |b| {
+    c.bench_function("bulk load baseline", move |b| {
         let points: Vec<_> = create_random_points(DEFAULT_BENCHMARK_TREE_SIZE, SEED_1);
 
         b.iter(|| {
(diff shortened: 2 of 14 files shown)


adamreichold:
Personally, I am very much open to bike-shedding the name as I think `GeomWithCachedEnvelope` is a bit of a mouthful. Maybe just `CachedEnvelope`? The `geom` method could also be dropped entirely due to the `Deref` impl.

rmanoka:
Nice, this is great.  Reg. the name:  we could also merge this func. into `GeomWithData<T::Envelope, T>` and add the `PointDist` impl.  We can then type alias it, and / or add a constructor `GeomWtihData::cached_envelope`.

Should we also add the explanation on why this is necessary somewhere more visible to most users?  Probably the `RTreeObject` trait or the main data-structure itself?

urschrei:
> Should we also add the explanation on why this is necessary somewhere more visible to most users? Probably the RTreeObject trait or the main data-structure itself?

Yes, definitely. From the perspective of someone looking for a spatial index, it's currently not even obvious what rstar is for; that is, you have to know (or find out) what an R*-tree is. A light edit of the docs to make it a little clearer what's going on, and when to choose RTreeObject, GeomWithData, GeomWithCachedEnvelope would go a long way I think. We can do that in a follow-up PR.

rmanoka (approved):

adamreichold:
> Nice, this is great. Reg. the name: we could also merge this func. into GeomWithData<T::Envelope, T> and add the PointDist impl. We can then type alias it, and / or add a constructor GeomWtihData::cached_envelope.

I admittedly do not follow. You mean using `GeomWithData` and store the envelope instead of the underlying geometry in `geom` and the geometry in `data`? How would the `PointDistance` impl decided whether to forward to `geom` or `data` in these circumstances? Wouldn't the impls overlap like

```console
error[E0119]: conflicting implementations of trait `object::PointDistance` for type `primitives::geom_with_data::GeomWithData<_, _>`
  --> rstar/src/primitives/geom_with_data.rs:69:1
   |
48 | impl<R: PointDistance, T> PointDistance for GeomWithData<R, T> {
   | -------------------------------------------------------------- first implementation here
...
69 | impl<T: PointDistance> PointDistance for GeomWithData<T::Envelope, T> {
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ conflicting implementation for `primitives::geom_with_data::GeomWithData<_, _>`
```

rmanoka:
Yes, I didn't think about the point dist implementation properly.  So let's go with what you have; may be a shorter name if you find one.

adamreichold:
> So let's go with what you have; may be a shorter name if you find one.

I'd propose just using `CachedEnvelope<T>`, i.e. dropping the `GeomWith` part and exclusively relying on `Deref` instead of a `geom` method. Agreed?

adamreichold:
Merging since the shorter name seems uncontroversial and the code itself was previously approved.

adamreichold:
Ouch, I totally forgot that we use bors here and merged manually for which I am sorry!

@urschrei I think it should be possible to prevent manual merges via the GitHub UI using the branch protection rules?

urschrei:
Oh don't worry! I'll have a look at the settings.

PR #122, opened 2023-04-23
Deprecate GeomWithData .inner() method in favour of Deref

urschrei:
This matches the CachedEnvelope impl

---


adamreichold:
Hhhmmm, I am admittedly not completely sure about this as `GeomWithData` is as much `Geom` as it is `Data`, so I don't think it is obvious that it should dereference into the geometry instead of the data part. Not sure whether this is really similar to the `CachedEnvelope` case where nothing is added conceptionally.

urschrei:
Hmm good point. Let's abandon this for now…

PR #125, opened 2023-04-27
Add polygon benchmarks

adamreichold:
---

We used those to test different approaches at memoizing envelopes before settling on the combinators as the choice without hidden costs so we did not need them in the end. But I think they would still be useful to have in the long run.


urschrei (approved):

adamreichold:
bors r=urschrei

PR #126, opened 2023-05-30
Prepare for 0.11.0 release

urschrei:
---

### DIFF
--- a/rstar/CHANGELOG.md
+++ b/rstar/CHANGELOG.md
@@ -1,4 +1,4 @@
-# UNRELEASED
+# 0.11.0
 
 ## Added
 - Add `CachedEnvelope` combinator which simplifies memoizing envelope computations. ([PR](https://github.com/georust/rstar/pull/118))
--- a/rstar/Cargo.toml
+++ b/rstar/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rstar"
-version = "0.10.0"
+version = "0.11.0"
 authors = ["Stefan Altmayer <stoeoef@gmail.com>", "The Georust Developers <mods@georust.org>"]
 description = "R*-tree library for the Rust ecosystem"
 documentation = "https://docs.rs/rstar/"


adamreichold (approved):

rstar/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "rstar"
-version = "0.10.0"
+version = "0.11.0"
```
michaelkirk, line 3:
```suggestion
version = "0.10.1"
```

rstar/CHANGELOG.md
```diff
@@ -1,4 +1,4 @@
-# UNRELEASED
+# 0.11.0
```
michaelkirk, line 1:
```suggestion
# 0.10.1
```

michaelkirk (commented):
I *think* these are all non-breaking changes, right?

adamreichold:
> I _think_ these are all non-breaking changes, right?

Technically, yes. But we did bump our MSRV. Personally, I find it impolite to do this in a patch release.

urschrei:
@michaelkirk Yes my bad, but I think Adam makes a good point about the MSRV

rmanoka:
Changes lgtm;  agree with bumping to `0.11.0` as some compilation might break elsewhere.

rmanoka (approved):

michaelkirk (approved):
The "should MSRV be breaking" debate is pretty well trodden territory. I'm in the "it does more harm than good" camp, but I won't die on that hill this time.

I will note that this change brings us past the introduction of `rust-version` (introduced in 1.56), so future bumps to `rust-version` have no case for being a breaking semver bump as I understand it.

michaelkirk (approved):
The "should MSRV be breaking" debate is pretty well trodden territory. I'm in the "it does more harm than good" camp, but I won't die on that hill this time.

I will note that this change brings us past the introduction of `rust-version` (introduced in 1.56), so future bumps to `rust-version` have no case for being a breaking semver bump as I understand it.

urschrei:
> The "should MSRV be breaking" debate is pretty well trodden territory. I'm in the "it does more harm than good" camp, but I won't die on that hill this time.
> 
> I will note that this change brings us past the introduction of `rust-version` (introduced in 1.56), so future bumps to `rust-version` have no case for being a breaking semver bump as I understand it.

I for one am delighted that we can settle this issue using metadata.

PR #127, opened 2023-06-13
Fix a stack overflow error in DrainIterator

nickguletskii:
---

Fixes a stack overflow error when using `RTree::drain_with_selection_function` or any other method that uses `DrainIterator`.

### DIFF
--- a/rstar/CHANGELOG.md
+++ b/rstar/CHANGELOG.md
@@ -1,3 +1,8 @@
+# Unreleased
+
+## Changed
+- Fixed a stack overflow error in `DrainIterator::next`
+
 # 0.11.0
 
 ## Added
--- a/rstar/src/algorithm/removal.rs
+++ b/rstar/src/algorithm/removal.rs
@@ -129,7 +129,7 @@ where
     type Item = T;
 
     fn next(&mut self) -> Option<Self::Item> {
-        loop {
+        'attempt_loop: loop {
             // Get reference to top node or return None.
             let (node, idx, remove_count) = match self.node_stack.last_mut() {
                 Some(node) => (&mut node.0, &mut node.1, &mut node.2),
@@ -149,7 +149,7 @@ where
                                 RTreeNode::Parent(node) => node,
                             };
                             self.node_stack.push((child, 0, 0));
-                            return self.next();
+                            continue 'attempt_loop;
                         }
                         RTreeNode::Leaf(ref leaf) => {
                             if self.removal_function.should_unpack_leaf(leaf) {


adamreichold (approved):

rmanoka:
Could we add a test-case to ensure it does not regress?  It's not clear what the bug was from the code diff.

nickguletskii:
> Could we add a test-case to ensure it does not regress? It's not clear what the bug was from the code diff.

I've added the quick-and-dirty fuzz test that I've implemented in order to track down this bug. Unfortunately, I don't have a simpler test case - it seems that this bug doesn't manifest itself that often.

adamreichold:
> > Could we add a test-case to ensure it does not regress? It's not clear what the bug was from the code diff.
> 
> I've added the quick-and-dirty fuzz test that I've implemented in order to track down this bug. Unfortunately, I don't have a simpler test case - it seems that this bug doesn't manifest itself that often.

I am not a fan of adding this library-specific fuzz test in this PR here. Especially since the change really is a quality of implementation issue, i.e. it is basically a manual tail call optimization.

I would like to include the fuzz test, but as a separate PR that can be reviewed without the time pressure of a bugfix release. For this PR, a single test case for example using a synthetic selection function would be fine for me, but I would also merge this without a test case triggering the stack overflow. (This depends on the environment in any case, i.e. I guess this is might be possible to trigger by running an existing test in a thread with an artificially small stack size set via [`thread::Builder::stack_size`](https://doc.rust-lang.org/stable/std/thread/struct.Builder.html#method.stack_size).)

nickguletskii:
> > > Could we add a test-case to ensure it does not regress? It's not clear what the bug was from the code diff.
> > 
> > 
> > I've added the quick-and-dirty fuzz test that I've implemented in order to track down this bug. Unfortunately, I don't have a simpler test case - it seems that this bug doesn't manifest itself that often.
> 
> I am not a fan of adding this library-specific fuzz test in this PR here. Especially since the change really is a quality of implementation issue, i.e. it is basically a manual tail call optimization.
> 
> I would like to include the fuzz test, but as a separate PR that can be reviewed without the time pressure of a bugfix release. For this PR, a single test case for example using a synthetic selection function would be fine for me, but I would also merge this without a test case triggering the stack overflow. (This depends on the environment in any case, i.e. I guess this is might be possible to trigger by running an existing test in a thread with an artificially small stack size set via [`thread::Builder::stack_size`](https://doc.rust-lang.org/stable/std/thread/struct.Builder.html#method.stack_size).)

Sorry for the delay again. I've removed the commits containing the fuzz test from this PR.

As for manually setting the stack size, I haven't had much luck reproducing this bug with the existing tests or tests derived from them even with a 64kb stack. I suspect that it might only exhibit itself after a lot of interleaved inserts and removals of random leaves with loose-fitting AABBs.

rmanoka:
Apologies, it took me a bit to understand that this is a tail-call optimization.  I'm okay with adding this without a test-case too.  Currently, I think we do not re-balance during removal, so our data-structure can get very deep (depth = O(nodes)) if used as a dynamic tree.

rmanoka (approved):

adamreichold:
bors r=adamreichold,rmanoka

Issue #28, opened 2020-03-22
Improve interoperability with other linear algebra crates


Stoeoef:
It would be nice if rstar had first class support for other geometry crates, e.g. `nalgebra` or `cgmath`.

Up to now, users either need to

- Implement their own point type (boiler plate / cannot easily use third-party point types due to orphan rule)
- Use the built in point type (like `[f64; 2]`, has only limited functionality)

One promising path forward might be adding support for the [mint crate](https://crates.io/crates/mint) crate. Adding `Point` impls for the mint types should allow to add them to an R-Tree directly. Both  `cgmath` and `nalgebra` support mint already.

This issue originated from #25

ripytide:
Could this be closed as fixed with #134 now merged?

adamreichold:
Agreed. Let's consider the `mint` support the canonical integration point as long as nobody asks for direct support of a particular crate like `nalgebra`.

PR #134, opened 2023-09-09
add mint feature for Point trait impl for mint::Point2 and mint::Point3

ripytide:
---

Helps with #28

### DIFF
--- a/rstar/src/mint.rs
+++ b/rstar/src/mint.rs
@@ -1,7 +1,7 @@
 //! [`mint`](https://crates.io/crates/mint) is a library for
 //! interoperability between maths crates, for example, you may want
 //! to use [nalgebra](https://crates.io/crates/nalgebra) types for
-//! representing your points and ALSO use the same points with the
+//! representing your points and _also_ use the same points with the
 //! `rstar` library.
 //!
 //! Here is an example of how you might do that using `mint` types for
--- a/rstar/src/mint.rs
+++ b/rstar/src/mint.rs
@@ -6,7 +6,9 @@
 //!
 //! Here is an example of how you might do that using `mint` types for
 //! compatibility between the two libraries. Make sure to enable the
-//! `mint` features on both `nalgebra` and `rstar` for this to work.
+//! `mint` features on both the `nalgebra` and the `rstar` crates for
+//! this to work. You will also need to depend on the
+//! [`mint`](https://crates.io/crates/mint) crate.
 //!
 //! ```
 //! use rstar::RTree;
--- a/rstar/src/mint.rs
+++ b/rstar/src/mint.rs
@@ -1,3 +1,33 @@
+//! [`mint`](https://crates.io/crates/mint) is a library for
+//! interoperability between maths crates, for example, you may want
+//! to use [nalgebra](https://crates.io/crates/nalgebra) types for
+//! representing your points and ALSO use the same points with the
+//! `rstar` library.
+//!
+//! Here is an example of how you might do that using `mint` types for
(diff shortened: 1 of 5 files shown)


adamreichold:
I think if we do this, we should aim for broader coverage, e.g. include at least also `mint::Point3`.

ripytide:
I've added an impl for mint::Point3

adamreichold:
I am sorry, for not asking for this earlier, but I do think we need at least some smoke test. Less for the actual code which is simple enough but for the integrate to actually work.

Could you add one of the crates supporting `mint` as a `dev-dependency` and write simple smoke tests for the two integrations, e.g. adding points using the foreign types to a tree and running some query on it? Thanks!

ripytide:
Unless the `rstar` library decided to use `impl Into<T> where T: Point` for accepting points in every interface in the lib which doesn't seem to be the case.

adamreichold:
Indeed `mint::Point2/3` are concrete types, but then that observation is basically the point: How would usage of this integration look in practice? What do I have to do to convert between this and the actual target types?

Actually, we might want to make these doctests to show case to readers of the documentation and ensure to the enable the `mint` dependency on docs.rs.

ripytide:
That's a good point, I should have probably given my use-case up front so help illustrate why `impl Point for mint::PointN` is useful.

In my use-case I am using a foreign point type, let's say `nalgebra::Point2` for example, I then want to store these in an `rtree:RTree` which gives me a few options since `nalgebra::Point2` rightfully does not implement `rtree::Point`:
1. Make a local newtype wrapper for `nalgebra::Point2` and manually implement `rtree::Point` on that newtype and then convert `nalgebra::Point2`s into my newtype before storing them in the `rtree::RTree`.
2. Convert my `nalgebra::Point2`s into `mint::Point2` (already implemented in the nalgebra crate with the "mint" feature) and then store those in the `rtree::RTree` (assuming this PR was merged so `mint::Point2: rtree::Point`)
3. Merge this PR and overhaul the API trait bounds [here](https://github.com/georust/rstar/pull/134#issuecomment-1713779963) to be able to store `nalgebra::Point2`s directly

Option 1 does not require changing the `rstar` library but is not very for users ergonomic as they have to make their own newtype everytime they want to store a foreign type.
Option 2 is more ergonomic as users can just store `mint::Point2`s and use the ready-to-go conversions with this PR
Option 3 would require an overhaul of the libraries trait bounds on most things which is a lot of library work but would allow users to store raw foreign types such as `nalgebra::Point2`. These impls in this PR would also be required for this method.

ripytide:
I am happy to add doc-tests for a use-case of Option 2 if you are happy to go that route

adamreichold:
Yes, option 2 was my understanding of how this is supposed to be used and I would be glad if we could add doctests with working examples e.g. using `nalgebra` as the integration target.

ripytide:
I've added documentation to the `mint` module with a doctest example of an integration with `nalgebra`. I also fixed a non-deterministic error with the `nearest_neighbors` that was failing for me.

rstar/src/mint.rs
```diff
@@ -0,0 +1,94 @@
+//! [`mint`](https://crates.io/crates/mint) is a library for
+//! interoperability between maths crates, for example, you may want
+//! to use [nalgebra](https://crates.io/crates/nalgebra) types for
+//! representing your points and ALSO use the same points with the
```
adamreichold, line 4: Please don't use all-caps to emphasize, a simple `_also_` should suffice.

adamreichold (approved):
Should be squash-merged and a nit on formatting the docs, otherwise LGTM.

PR #136, opened 2023-09-27
Better distance_2 docs

urschrei:
---

### DIFF
--- a/rstar/src/object.rs
+++ b/rstar/src/object.rs
@@ -145,7 +145,12 @@ pub trait RTreeObject {
 /// assert!(circle.contains_point(&[1.0, 0.0]));
 /// ```
 pub trait PointDistance: RTreeObject {
-    /// Returns the squared euclidean distance between an object to a point.
+    /// Returns the squared distance between an object to a point.
+    ///
+    /// # Notes
+    /// While euclidean distance will be the correct choice for most use cases, any distance metric
+    /// fulfilling the [usual axioms](https://en.wikipedia.org/wiki/Metric_space)
+    /// can be used when implementing this method.
     fn distance_2(
         &self,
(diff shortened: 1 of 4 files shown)


rstar/src/object.rs
```diff
@@ -145,7 +145,12 @@ pub trait RTreeObject {
 /// assert!(circle.contains_point(&[1.0, 0.0]));
 /// ```
 pub trait PointDistance: RTreeObject {
-    /// Returns the squared euclidean distance between an object to a point.
+    /// Returns the squared distance between an object to a point.
```
michaelkirk, line 148:
```suggestion
    /// Returns the squared distance between an object and a point.
```

Or alternatively: 

```
 /// Returns the squared distance from an object to a point.
```

(I know this wasn't your text originally 😉)

michaelkirk (approved):

adamreichold:
I think the `Envelope` trait has the same issue for its `distance_2` method.

urschrei:
@adamreichold Good catch. Is it worth cautioning implementers that the two trait method implementations should use the same metric? I.e. can you imagine a situation in which they might be different?

adamreichold:
> @adamreichold Good catch. Is it worth cautioning implementers that the two trait method implementations should use the same metric? I.e. can you imagine a situation in which they might be different?

From how it is used in the nearest neighbour search, I think these two implementations of the underlying must definitely agree as the distance from envelopes to points is used as lower bound for the distance between objects inside these envelopes to these points.

(From a semantics point of view, it would probably be nicer if the envelope type would implement `PointDistance` just as the object type does to stress this equivalence, but I guess this is not worth it to change retroactively.)

adamreichold (approved):

PR #137, opened 2023-10-08
Add cached envelope benchmark

urschrei:
---

### DIFF
--- a/rstar-benches/benches/benchmarks.rs
+++ b/rstar-benches/benches/benchmarks.rs
@@ -68,13 +68,9 @@ fn bulk_load_complex_geom_cached(c: &mut Criterion) {
     c.bench_function(
         "Bulk load complex geo-types geom with cached envelope",
         move |b| {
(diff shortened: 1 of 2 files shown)


adamreichold:
Not really part of the change but since we are touching this file:
- `circular_polygon` is `pub` but I don't think it needs to be.
- `create_random_polygons` could return `impl Iterator<Item=Polygon<f64>>` to avoid unnecessary intermediate collections.

urschrei:
@adamreichold like this?

adamreichold:
> @adamreichold like this?

Almost. The new `bulk_load_complex_geom_cached` benchmark still collects and then maps instead immediately doing something like

```rust
let cached: Vec<_> =
    create_random_polygons(DEFAULT_BENCHMARK_TREE_SIZE, 4096, SEED_1)
        .map(CachedEnvelope::new)
        .collect();
```

adamreichold (approved):

Issue #142, opened 2023-12-20
Enable merge queues and CI runs for PR


adamreichold:
Since the public instance of bors-ng is now offline, we are lacking any pre-merge CI checks for pull requests. We should at least extend our CI to also test pull requests in addition to local branches. Further, we might want to enable GitHub merge queues to get something close to the previous approach using bors-ng.

PR #143, opened 2023-12-20
Update CI as a precursor to enabling merge queues

urschrei:
See #142

---

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -48,11 +48,14 @@ jobs:
     strategy:
       matrix:
         container_image:
+          # We aim to support rust-stable plus (at least) the prior 3 releases,
+          # giving us about 6 months of coverage.
+          #
           # Minimum supported rust version (MSRV)
-          - "georust/geo-ci:rust-1.65"
+          - "georust/geo-ci:proj-9.2.1-rust-1.65"
           # Two most recent releases - we omit older ones for expedient CI
-          - "georust/geo-ci:rust-1.71"
-          - "georust/geo-ci:rust-1.72"
+          - "georust/geo-ci:proj-9.2.1-rust-1.71"
+          - "georust/geo-ci:proj-9.2.1-rust-1.72"
     container:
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -1,26 +1,41 @@
...
     needs:
       - rstar
       - check
       - no_std
+    if: success()
     steps:
       - name: Mark the job as a success
-        if: success()
         run: exit 0
+  ci-failure:
```
adamreichold, line 29: It is somewhat reasonable, but just to have the alternative discussed here, we could also have a single job dynamically checking the other jobs results, e.g. in PyO3 we use

```yaml
conclusion:
needs:
    - fmt
    - check-msrv
    - clippy
    - build-pr
    - build-full
    - valgrind
    - careful
    - coverage
    - emscripten
if: always()
runs-on: ubuntu-latest
steps:
    - name: Result
    run: |
        jq -C <<< "${needs}"
        # Check if all needs were successful or skipped.
        "$(jq -r 'all(.result as $result | (["success", "skipped"] | contains([$result])))' <<< "${needs}")"
    env:
        needs: ${{ toJson(needs) }}
```

adamreichold: Alternatively, we could at least factor out the job list using a YAML anchor?

adamreichold: Certainly more DRY.

adamreichold (approved):

adamreichold:
Did you configure the merge queue for squash merging? Otherwise, I would suggest manually squashing this PR before merging.

urschrei:
> Did you configure the merge queue for squash merging? Otherwise, I would suggest manually squashing this PR before merging.

Configured to "squash and merge"

urschrei:
Hum. That shouldn't have merged before the tests passed

urschrei:
> Hum. That shouldn't have merged before the tests passed

Ah forgot to update setting with new ci check name. Now needs `conclusion` to pass before merging.

PR #154, opened 2024-01-14
Use SmallHeap also for the iterator-based nearest neighbour search and optimize its spill implementation.

adamreichold:


urschrei (approved):

urschrei:
@adamreichold Are you happy for this to merge so it can be included in the new release?

adamreichold:
Yes, I don't plan on changing anything here.

(From measurements, I suspect we could increase the inline capacity from 32 to 64, but I only tested this using random trees and hence are not confident enough to commit to this yet.)

PR #160, opened 2024-03-14
Use unstable sort for envelopes and node reinsertion

grovesNL:
---

I was just looking at some unrelated AABB performance testing and noticed stable sort being used in a couple places.

I don't have a great performance/memory profile to show a meaningful difference here, but there doesn't seem to be a reason we need to stable sort in these places.

(probably not worth a changelog entry)

### DIFF
--- a/rstar/CHANGELOG.md
+++ b/rstar/CHANGELOG.md
@@ -1,3 +1,8 @@
+# Unreleased
+
+## Changed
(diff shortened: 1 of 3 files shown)


adamreichold (approved):
Agree that there is no reason to use a stable sort here.

The bulk loading for example already uses unstable selection when it needs to sort envelopes.

PR #162, opened 2024-04-06
Fix overflows applying selection iterators to empty trees by choosing a more tame value for AABB::new_empty

### LINKED ISSUE #161: RTree panics when running locate_within_distance on an empty tree
I am creating an application in which a loop runs. During this loop a new rtree is created, and potentially something is inserted, then I run locate_within_distance on the resulting tree. The issue is that if nothing gets inserted at all, the rtree remains empty, then running locate_within_distance() panics. Here is some code to reproduce the panic

```
fn main() {

    let mut empty_tree : RTree<[i64;3]> = RTree::new();
   
    let all = empty_tree.locate_within_distance([0,0,0], 10);
}

```

adamreichold:
---

Closes #161

### DIFF
--- a/rstar/CHANGELOG.md
+++ b/rstar/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Changed
 - Switched to unstable sort for envelopes and node reinsertion ([PR](https://github.com/georust/rstar/pull/160))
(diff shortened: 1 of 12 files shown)


urschrei (approved):

gold-silver-copper:
Thank you it works great now !

PR #164, opened 2024-04-24
RFC: Provide selection methods based on internal iteration

### LINKED ISSUE #163: Avoiding `SmallVec` in `SelectionIterator`
In `SelectionIterator` and `SelectionIteratorMut` we use a `SmallVec` to queue nodes as we're iterating over the nodes in the tree, e.g.:
https://github.com/georust/rstar/blob/84d12654104e783011f24267145fb6bfccd2a30e/rstar/src/algorithm/iterators.rs#L46

Constantly resizing this `SmallVec` seems to show up on some of my profiles where I have relatively dense trees.

I was wondering if there might be a way to avoid `SmallVec` here somehow but it's not clear how it could work. Maybe some other crates already have found a good pattern to avoid this kind of iteration queue? e.g., if we could somehow represent a flattened position of the next node then maybe that could work, but I don't know if this would be possible without changing how nodes are stored.

adamreichold:
---

This avoids the overhead of allocating an internal buffer to keep track of upcoming nodes when implementing the `Iterator` trait.

I also found a mistake in the old code from #37 (lack of early return in the parent case) and now the benchmarks also look somewhat nicer, i.e. directly comparing internal and external iteration on the same data set:

```console
locate_at_point (successful)
                        time:   [115.62 ns 116.51 ns 117.43 ns]

locate_at_point_int (successful)
                        time:   [66.831 ns 67.264 ns 67.653 ns]

                        time:   [167.70 ns 168.03 ns 168.34 ns]

locate_at_point_int (unsuccessful)
                        time:   [167.90 ns 168.28 ns 168.64 ns]
```

Closes #163

### DIFF
--- a/rstar-benches/Cargo.toml
+++ b/rstar-benches/Cargo.toml
@@ -4,11 +4,11 @@ version = "0.1.1"
 authors = ["Stefan Altmayer <stoeoef@gmail.com>", "The Georust Developers <mods@georust.org>"]
 
 [dev-dependencies]
-criterion = { version = "0.4.0", features = ["html_reports"] }
-geo = "0.26.0"
-geo-types = { version = "0.7.9", features = ["use-rstar_0_10"] }
-rand = "0.7"
-rand_hc = "0.2"
+criterion = { version = "0.5.0", features = ["html_reports"] }
+geo = "0.28.0"
+geo-types = { version = "0.7.9", features = ["use-rstar_0_12"] }
+rand = "0.8"
+rand_hc = "0.3"
 rstar = { path = "../rstar" }
 
 [[bench]]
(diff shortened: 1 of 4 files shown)


rmanoka:
I'm for having this recursive impl. as an alternative to the iterator approach.

urschrei:
@adamreichold This is ready to merge, right? The only comment I have is that it would be good to have some brief documentation about when use of the new methods is preferable.

adamreichold:
> @adamreichold This is ready to merge, right? The only comment I have is that it would be good to have some brief documentation about when use of the new methods is preferable.

Yes, this is ready to go from my part. I can add the requested documentation, but do you have a preference where to put this? On the `RTree` type discussing both approaches? On the new methods to contrast them with the existing ones?

adamreichold:
I added some documentation on the type level discussing the selection methods and their alternative implementations using internal iteration.

PR #165, opened 2024-04-28
Update smallvec

dabreegster:
I'm having build conflicts when I include both `rstar` and another crate using new `smallvec`. Any reason the version specified in rstar is so old? Tests pass just fine with this change, but I haven't looked at any benchmarks yet. May also be related to #163.

### DIFF
--- a/rstar/Cargo.toml
+++ b/rstar/Cargo.toml
@@ -19,7 +19,7 @@ maintenance = { status = "actively-developed" }
 heapless = "0.8"
 num-traits = { version = "0.2", default-features = false, features = ["libm"] }
 serde = { version = "1.0", optional = true, default-features = false, features = ["alloc", "derive"] }
-smallvec = "1.6"
+smallvec = "1.13"
 mint = { version = "0.5.9", optional = true }
 
 [features]


adamreichold:
>  Any reason the version specified in rstar is so old?

The change you made should not influence version selection in a downstream project as 1.6 and 1.13 are semver-compatible. Meaning even if we have `1.6` meaning `^1.6` meaning `>=1.6,<2`  here, `cargo update` should still result in your project using 1.13.

So I fear your build conflicts have a different root cause. Maybe you can share the error message here and we can try to figure it out?

dabreegster:
Hmm, indeed; I forgot that "1.6" means "^1.6" and not "=1.6".

I'm depending on rstar 0.12 and a crate on gitlab, which has this dependency on `smallvec`: `smallvec = { version = "1.13", default-features = false, optional = true }`
 https://gitlab.com/LeLuxNet/Muv/-/blob/main/muv-osm/Cargo.toml?ref_type=heads#L17

The error:

```
Error: `cargo metadata` exited with an error:     Updating git repository `https://gitlab.com/LeLuxNet/Muv`
    Updating crates.io index
error: failed to select a version for `smallvec`.
    ... required by package `muv-osm v0.1.0 (https://gitlab.com/LeLuxNet/Muv#38eddd4b)`
    ... which satisfies git dependency `muv-osm` of package `backend v0.1.0 (/home/dabreegster/15m/backend)`
versions that meet the requirements `^1.13` are: 1.13.2, 1.13.1

all possible versions conflict with previously selected packages.

  previously selected package `smallvec v1.11.2`
    ... which satisfies dependency `smallvec = "^1.6"` (locked to 1.11.2) of package `rstar v0.12.0`
    ... which satisfies dependency `rstar = "^0.12.0"` (locked to 0.12.0) of package `backend v0.1.0 (/home/dabreegster/15m/backend)`

failed to select a version for `smallvec` which could resolve this conflict

Caused by: `cargo metadata` exited with an error:     Updating git repository `https://gitlab.com/LeLuxNet/Muv`
    Updating crates.io index
error: failed to select a version for `smallvec`.
    ... required by package `muv-osm v0.1.0 (https://gitlab.com/LeLuxNet/Muv#38eddd4b)`
    ... which satisfies git dependency `muv-osm` of package `backend v0.1.0 (/home/dabreegster/15m/backend)`
versions that meet the requirements `^1.13` are: 1.13.2, 1.13.1

all possible versions conflict with previously selected packages.

  previously selected package `smallvec v1.11.2`
    ... which satisfies dependency `smallvec = "^1.6"` (locked to 1.11.2) of package `rstar v0.12.0`
    ... which satisfies dependency `rstar = "^0.12.0"` (locked to 0.12.0) of package `backend v0.1.0 (/home/dabreegster/15m/backend)`

failed to select a version for `smallvec` which could resolve this conflict
```

... Most oddly, if I `cargo update -p smallvec` and then add the Muv dependency, everything is totally fine. I expected the dependency resolver to understand smallvec was upgradeable. Looks like this is some possible bug in cargo; I'll go investigate there. Thanks for your help!

adamreichold:
> Looks like this is some possible bug in cargo; I'll go investigate there.

From the error message, I wonder if `cargo metdata` was given the `--locked` flag or else if `cargo metadata` is allowed to update the dependency resolution at all?

dabreegster:
> I wonder if cargo metdata was given the --locked flag 

This might be it! I was using `wasm-pack`, and I could see a strong argument for `--locked` being a default there

Issue #111, opened 2023-02-11
Question: Why there is parameters verification that point dimension must be at least 2?


ytskuh:
I would like to create a R*-tree for points that could have a single dimension, but it says "thread 'main' panicked at 'Point dimension too small - must be at least 2'"

urschrei:
Could you explain more about what you're trying to achieve? Preferably with a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example). I mostly work with geometries in R2, so I have to admit that I simply don't understand the phrase "points that could have a single dimension".

michaelkirk:
I'm not sure it makes sense to use an RTree for less than 2 dimensions.

If I wanted to efficiently insert and query over a 1 dimensional range, I think I'd use a btree:

https://doc.rust-lang.org/stable/std/collections/struct.BTreeMap.html#method.range

Or do you mean that your points will be a mix of 1 and 2(+) dimensions?

ytskuh:
I am working with a program that works with problems could be set in different dimensions - I just want to solve all situation with the same piece of code. It's right that btree or other solution is better for one dimensional problem. And I have found the crate that is more suitable for me (kd_tree)

grovesNL:
@urschrei I have a use case where I currently use a `BTreeMap<u32, u32>` to represent intervals (key as the start, value as the end). It's really read heavy because I do a lot of range queries over it to see which intervals overlap. I think a one-dimensional RTree could be a better alternative for this use case and I'm currently attempting to do performance testing, but ran into this point dimension limitation. I'm going to work through it on a fork so I can test it out anyway, but just wanted to mention my use case here.

@michaelkirk `BTreeMap` is good but there are optimizations you can apply with a 1D bounding box/envelope that wouldn't apply to a generic BTree. My intuition is that a 1D r*-tree could perform better, but I haven't confirmed this yet.

PR #169, opened 2024-06-27
Allow 1D `RTree`s

grovesNL:
---

This PR removes the parameter validation to allow 1D `RTree`s to be created.

1D `RTree`s already seem to work fine as far as I can tell from a simple example I just tested. I've looked through the places where we reference "dimensions" but most of them just loop from `0..dimensions`. We even explicitly support single dimensions in 1D tuples and some other places.

Motivation described in #111

### DIFF
--- a/rstar/CHANGELOG.md
+++ b/rstar/CHANGELOG.md
@@ -4,6 +4,7 @@
 - Switched to unstable sort for envelopes and node reinsertion ([PR](https://github.com/georust/rstar/pull/160))
 - Use a more tame value for `AABB::new_empty` to avoid overflow panics applying selections on empty trees ([PR](https://github.com/georust/rstar/pull/162))
 - Avoid infinite recursion due to numerical instability when computing the number of clusters ([PR](https://github.com/georust/rstar/pull/166))
+- Allow 1D `RTree`s to be created ([PR](https://github.com/georust/rstar/pull/169))
 
 # 0.12.0
 
--- a/rstar/src/params.rs
+++ b/rstar/src/params.rs
@@ -1,5 +1,5 @@
 use crate::algorithm::rstar::RStarInsertionStrategy;
-use crate::{Envelope, Point, RTree, RTreeObject};
+use crate::{RTree, RTreeObject};
 
 /// Defines static parameters for an r-tree.
 ///
@@ -106,11 +106,5 @@ pub fn verify_parameters<T: RTreeObject, P: RTreeParams>() {
         P::REINSERTION_COUNT < max_reinsertion_count,
         "REINSERTION_COUNT too large. Must be smaller than {:?}",
         max_reinsertion_count
-    );
-
-    let dimension = <T::Envelope as Envelope>::Point::DIMENSIONS;
-    assert!(
-        dimension > 1,
-        "Point dimension too small - must be at least 2"
     );
 }
--- a/rstar/src/point.rs
+++ b/rstar/src/point.rs
@@ -409,7 +409,7 @@ mod tests {
 
     macro_rules! test_tuple_configuration {
         ($($index:expr),*) => {
-            let a = ($($index),*);
+            let a = ($($index),*,);
             $(assert_eq!(a.nth($index), $index));*
         }
     }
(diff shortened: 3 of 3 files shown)


michaelkirk (approved):
LGTM. 

Did you notice any performance improvements vs. using `BTreeSet`?

michaelkirk:
> Edit: actually it looks like there are some places where it's returning false positives, I'll take a look

Ah, I just saw your edit.

adamreichold:
I don't think the point of not allowing one-dimensional geometry that it would not work, but rather that R trees are considerably over-engineered for the one-dimensional problem. I think the closest analogue is an interval tree which we for example use for time range queries (where we use an R tree for spatial queries).

I do think that all operations we provide here collapse to the interval tree version for the one-dimensional case, but it seems questionable to use this instead of providing a proper interval tree structure. Consider for example distance computations where we would uselessly compute squares or sometimes even worse square roots when we need the distance instead of the squared distance which would just collapse to computing the absolute value for the interval case.

The main conceptual difference here I think is that basically all nuances around R trees come from having to chose a single dimension each time the search space is to split into smaller parts. A problem which is completely absent when using a single dimension.

grovesNL:
@adamreichold Agreed, I'd like to use it for the exact use case of implementing an interval tree. I've evaluated most of the existing interval trees/range maps/segment tree crates but haven't been able to find anything suitable (usually cache-inefficient or inflexible).

My thought was that we could try to specialize the 1D `RTree` with traits as much as possible (beyond this initial support) to avoid some of the distance tests we need for higher dimensions. It also is totally fair to say we don't want support that complexity here and keep it restricted it to 2D or higher.

`RTree` already has most of the operations I'd like for the 1D case, so I could copy most of the implementation and remove all of the higher dimension functionality to specialize it, but I was hoping there was a way to share most of the implementation.

adamreichold:
> My thought was that we could try to specialize the 1D RTree with traits as much as possible (beyond this initial support) to avoid some of the distance tests we need for higher dimensions. It's also totally fair to say we don't want support that complexity here and keep it restricted it to 2D or higher.

I ran into this as well and my take away was that implementing the one-dimensional case from the ground up as an interval is indeed simpler than working to specialize the n-dimensional case trying to achieve a similar efficiency.

(I maintain my own little interval tree called `sif-itree` so I can memory map it directly, so for me that did solve the flexibility issue because I can implement exactly what I need which is just range queries. I guess PR to improve existing crates if you do not have the memory mapping use case would also be viable. I started out along those lines for example [here](https://github.com/main--/rust-intervaltree/pull/12).)

grovesNL:
I don't think it would be too difficult to specialize 1D here, but I'm ok with writing my own outside `rstar`.

For what it's worth I'd also like to eventually specialize 2D calculations too (e.g., SIMD overlap tests on a specialized 2D-only AABB) while sharing the same tree structure, selection iterators, etc.

adamreichold:
> For what it's worth I'd also like to eventually specialize 2D calculations too (e.g., SIMD overlap tests on a specialized 2D-only AABB) while sharing the same tree structure, selection iterators, etc.

I think this should be possible by reimplementing the traits defined by this crate and in the 2D case this should also not impose any infrastructure cost that is not actually necessary, in contrast to the one-dimensional case. (I also think we could carry those two-dimensional reimplementations upstream here because 2D is likely a very common case in GIS.)

grovesNL:
It looks like the false positives I noticed were due the `from_points` issue I split off into #170, so this seems to be working fine for 1D.

```rust
#[derive(Clone, Debug, PartialEq)]
#[repr(transparent)]
struct IntervalAabb(AABB<(i64,)>);

impl RTreeObject for IntervalAabb {
    type Envelope = AABB<IntervalPoint>;

        self.0
    }
}


let envelopes = vec![
    IntervalAabb(AABB::from_corners((3,), (4,))),
    IntervalAabb(AABB::from_corners((0,), (1,))),
    IntervalAabb(AABB::from_corners((10,), (12,))),
];

let tree = RTree::bulk_load(envelopes);

for x in tree.locate_in_envelope_intersecting(&AABB::from_corners((4,), (11,))) {
    println!("{:?}", x);
}

// Outputs:
// IntervalAabb(AABB { lower: (10,), upper: (12,) })
// IntervalAabb(AABB { lower: (3,), upper: (4,) })
```

Happy to close this PR if we'd prefer to keep 1D disabled. Either way it looks like `rstar` is already set up to handle 1D so it might be nice to expose it for use cases like this.

PR #182, opened 2024-11-05
Better handling of empty rects

michaelkirk:
- [ not yet] I added an entry to `rstar/CHANGELOG.md` if knowledge of this change could be valuable to users.
---

As an alternative to https://github.com/georust/rstar/pull/181, which fixes the same bug, but maintains the spirit of #162 by avoiding comparison with huge/tiny floats.

This is a draft because there are several things I'm not sure how to handle, and I'm out of time for today! But if someone is interested in building off of it, feel free!

### DIFF
--- a/rstar/src/aabb.rs
+++ b/rstar/src/aabb.rs
@@ -97,6 +97,11 @@ where
             self.min_point(point).sub(point).length_2()
         }
     }
+
+    fn is_empty(&self) -> bool {
+        self.lower
+            .all_component_wise(&self.upper, |min, max| min > max)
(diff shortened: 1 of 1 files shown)


adamreichold:
Having thought about this a bit, I think the max/min representation of empty AABB is indeed superior and avoids additional overhead for most users (i.e. those not using fixed-precision integer coordinates), i.e. I would prefer #181 or #184 over this one.

michaelkirk:
Superseded by https://github.com/georust/rstar/pull/184

PR #184, opened 2024-11-05
Revert back to min/max representation of empty AABB

### LINKED ISSUE #183: Performance degradation following 0.12.1 release
After upgrading the Rstar crate from version 0.12 to 0.12.1, I've experienced a dramatic slowdown in performance when using `.nearest_neighbor` in my public transit graph application. The operation, which connects all transit stops to the nearest street nodes, has slowed from ~50ms to over 30 seconds. All geometries are f64 `geo-types` points.

-- Expected Performance: ~2-5μs per `.nearest_neighbor` call (as observed with version 0.12)
-- Observed Performance: 10-20ms per `.nearest_neighbor` call with version 0.12.1
-- Change: Only the Rstar crate version was updated (from 0.12 to 0.12.1).

Cargo Flamegraph profiling didn’t pinpoint the issue due to heavy interference from Polars-related calls.
flamegraph_0_12

```rust
pub type IndexedPoint = GeomWithData<Point, Option<NodeIndex>>;

fn find_nearest_point_and_calculate_distance(
    point: &Point,
    tree: &RTree<IndexedPoint>,
) -> Result<(NodeIndex, f64), Error> {
    let instant = std::time::Instant::now();
    if let Some(nearest_point) = tree.nearest_neighbor(point) {
        println!("This took {:?}", instant.elapsed());
        let distance = Haversine::distance(*point, *nearest_point.geom()) / WALK_SPEED;
        let node = nearest_point.data.ok_or_else(|| {
            Error::NodeNotFound(format!("Nearest node not found for point {point:?}"))
        })?;
        Ok((node, distance))
    } else {
        Err(Error::NodeNotFound(format!(
            "Nearest node not found for point {point:?}"
        )))
    }
}

pub(crate) fn build_rtree(graph: &DiGraph<GraphNode, GraphEdge>) -> RTree<IndexedPoint> {
    let index_geo_vec: Vec<IndexedPoint> = graph
        .node_indices()
        .map(|node| {
            let node_data = graph.node_weight(node).unwrap();
            let node_point: Point = *node_data.geometry();
            IndexedPoint::new(node_point, Some(node))
        })
        .collect();

    RTree::bulk_load(index_geo_vec)
}

/// Connects Transit nodes (stops) to the nearest walk nodes.
pub(crate) fn connect_stops_to_streets(graph: &mut TransitGraph) -> Result<(), Error> {
    let instant = std::time::Instant::now();
    let rtree: RTree<IndexedPoint> = graph.rtree_ref().unwrap().clone();

    for node in graph.node_indices() {
        // check if there is already a transfer edge
        // This is required to avoid creating duplicate transfer edges
        // when merging multiple transit graphs on top of main street graph
        // (Work in progress)
        if graph
            .edges(node)
            .any(|edge| matches!(edge.weight(), GraphEdge::Transfer(_)))
        {
            continue;
        }

        let weight = graph
            .node_weight(node)
            .ok_or_else(|| Error::MissingValue("Node weight not found".to_string()))?;

        if let GraphNode::Transit(_) = weight {
            if let Ok((nearest_point_index, distance)) =
                find_nearest_point_and_calculate_distance(weight.geometry(), &rtree)
            {
                let edge = GraphEdge::Transfer(WalkEdge {
                    edge_weight: distance,
                });

                graph.add_edge(node, nearest_point_index, edge.clone());
                graph.add_edge(nearest_point_index, node, edge);
            }
        }
    }
    println!("This took {:?}", instant.elapsed());
    Ok(())
}
```

adamreichold:
---

This reverts back to the min/max representation of empty AABB due to regressions introduced by the numerically more tame but not invariant for merging representation.

To avoid regressing #161, it also adds a single check to avoid computing the distance to an empty envelope (of the root node of an empty tree). Integer coordinates are always prone to overflow but in the empty case we are forcing this onto the caller whereas every non-empty tree will have AABB that the caller supplied and where we can reasonably ask them to control for overflow or use a custom `Point` impl based on saturating arithmetic.

Fixes #183

### DIFF
--- a/rstar-benches/Cargo.toml
+++ b/rstar-benches/Cargo.toml
@@ -1,5 +1,6 @@
 [package]
 name = "rstar-benches"
+edition = "2015"
 version = "0.1.1"
 authors = ["Stefan Altmayer <stoeoef@gmail.com>", "The Georust Developers <mods@georust.org>"]
 
--- a/rstar/CHANGELOG.md
+++ b/rstar/CHANGELOG.md
@@ -1,3 +1,8 @@
+# Unreleased
+
+## Changed
+- Reverted the change to `AABB::new_empty` while still avoiding overflow panics applying selections on empty trees ([PR](https://github.com/georust/rstar/pull/184)
+
 # 0.12.1
 
 ## Added
--- a/rstar/src/aabb.rs
+++ b/rstar/src/aabb.rs
@@ -106,7 +106,12 @@ where
     type Point = P;
 
     fn new_empty() -> Self {
-        new_empty()
+        let max = P::Scalar::max_value();
+        let min = P::Scalar::min_value();
+        Self {
+            lower: P::from_value(max),
+            upper: P::from_value(min),
+        }
     }
 
     fn contains_point(&self, point: &P) -> bool {
@@ -219,20 +224,32 @@ where
     }
 }
 
-fn new_empty<P: Point>() -> AABB<P> {
-    let one = P::Scalar::one();
-    let zero = P::Scalar::zero();
-    AABB {
-        lower: P::from_value(one),
-        upper: P::from_value(zero),
-    }
-}
-
 #[cfg(test)]
 mod test {
     use super::AABB;
     use crate::envelope::Envelope;
     use crate::object::PointDistance;
+
+    #[test]
+    fn empty_rect() {
+        let empty = AABB::<[f32; 2]>::new_empty();
+
+        let other = AABB::from_corners([1.0, 1.0], [1.0, 1.0]);
+        let subject = empty.merged(&other);
+        assert_eq!(other, subject);
+
+        let other = AABB::from_corners([0.0, 0.0], [0.0, 0.0]);
+        let subject = empty.merged(&other);
+        assert_eq!(other, subject);
+
+        let other = AABB::from_corners([0.5, 0.5], [0.5, 0.5]);
+        let subject = empty.merged(&other);
+        assert_eq!(other, subject);
+
+        let other = AABB::from_corners([-0.5, -0.5], [-0.5, -0.5]);
+        let subject = empty.merged(&other);
+        assert_eq!(other, subject);
+    }
 
     /// Test that min_max_dist_2 is identical to distance_2 for the equivalent
     /// min max corner of the AABB. This is necessary to prevent optimizations
--- a/rstar/src/algorithm/iterators.rs
+++ b/rstar/src/algorithm/iterators.rs
@@ -53,11 +53,12 @@ where
     Func: SelectionFunction<T>,
 {
     pub(crate) fn new(root: &'a ParentNode<T>, func: Func) -> Self {
-        let current_nodes = if func.should_unpack_parent(&root.envelope()) {
-            root.children.iter().collect()
-        } else {
-            SmallVec::new()
-        };
+        let current_nodes =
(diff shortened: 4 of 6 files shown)


adamreichold:
This would be my personally preferred alternative to #181 and #182 trying to keep #161 working.

Generally speaking, I am really unsure whether the problems caused by fixed-precision integer coordinates are worth it, but this fix seems to be rather targeted and should not add measurable costs to the majority of floating-point coordinate users.

adamreichold:
(Whatever choice we make, we should probably yank 0.12.1 after releasing 0.12.2 to crates.io.)

urschrei:
OK – shall we merge this, release a patch version, and yank?

michaelkirk (approved):
Seems reasonable!

adamreichold:
> OK – shall we merge this, release a patch version, and yank?

If there are no objections from elsewhere, yes, I think that would be best to provide a quick resolution for affected projects.

I think the risk of an error in the fix is small for most users (as we are going back to the previous behaviour) and we'll just have to deal with any fallout from fixed-precision integer coordinate users. I tried to avoid it but I am admittedly not confident that it can be avoided.

urschrei:
0.12.1 is yanked

adamreichold:
> 0.12.1 is yanked

I think it is customary to yank only after an upgrade is available. Otherwise, the degree of eyebrow raising can be problematic. But I suspect you'll push 0.12.2 soon as well?

urschrei:
Released!

michaelkirk:
> I think it is customary to yank only after an upgrade is available.

At first I thought this made sense, but now I'm not sure. 

Assuming the fix is patch compatible (0.0.x++), if you wait to have the fix published before running `cargo yank`, it seems like the only practical consequence of `cargo yank` is to annotate what's shown on crates.io (and other indexes).

Screenshot 2024-11-05 at 14 09 07

If you already have the patch-compatible version published, `cargo update` will update to that fixed version, skipping the broken one anyway, right?

So it can be used to *prevent* people from updating. But it's not useful for "skipping" the broken one, since cargo already installs the most recent semver compatible one when running `cargo update`.

michaelkirk:
To summarize, it's my conclusion that `cargo yank` should be used *precisely* in the time after you know you have a problem, but before a fix is available.

urschrei:
Delightfully, we can unyank.

adamreichold:
A lot of CI runs tools like `cargo audit` or `cargo deny` which will warn about yanked dependencies. Warnings which are not actionable because no semver-compatible replacement is available is what I was referring to. Admittedly, this is not so bad if an earlier patch release is available because you can most likely downgrade successfully, but for example yanking 0.12.0 is significantly worse.

adamreichold:
> Delightfully, we can unyank.

I would not recommend that due to as written, `cargo update` is not the only concern. Yanking sends a relevant signal to different tools and also users with existing locked dependencies.

michaelkirk:
>  It's my conclusion that cargo yank should be used precisely in the time after you know you have a problem, but before a fix is available.

I phrased this poorly — In case there was confusion, I'm not recommending `unyank`ing in this case.

I'll again:

You should yank as soon as you know there is a problem, to keep people from updating to the broken version, even before a fix is available, ensuring legacy users stay on the old version until the fix is available.

`yanking` after a fix is published isn't *harmful* it's just ~~not very useful~~.

As @adamreichold points out, there is some utility to yanking, even if you've already shipped the fix.

PR #189, opened 2024-12-16
Make RTree methods take owned values where it makes sense

### LINKED ISSUE #188: Functions needlessly taking envelope by reference
A lot of functions currently take a `T::Envelope` by reference, and then immediately clones it. For example `locate_in_envelope`: https://github.com/georust/rstar/blob/c1b704de7606676fbc871badfde66bf10fd10536/rstar/src/rtree.rs#L336

This is both unnecessary and makes it hard to make helper functions that return iterators using these functions. In my current project I want a function to calculate an AABB and return a `Box<dyn Iterator<Item = &SomeType>>`. But because the envelope is taken by reference I now need to also store the AABB with the same lifetime.

Instead I propose that these functions take ownership of the envelope.

kristoff3r:
---

Fixes #188.

This makes the API nicer, and I don't think I touched anything low-level enough for it to have performance implications.

### DIFF
--- a/rstar/CHANGELOG.md
+++ b/rstar/CHANGELOG.md
@@ -3,6 +3,9 @@
 ## Added
 - Added missing re-exports of nearest neighbor iterators so they can be named in downstream crates. ([PR](https://github.com/georust/rstar/pull/186))
 
+## Changed
+- Made `RStar` methods take `Point` and `Envelope` as owned values where it makes sense ([PR](https://github.com/georust/rstar/pull/189))
+
 # 0.12.2
 
 ## Changed
--- a/rstar/src/rtree.rs
+++ b/rstar/src/rtree.rs
@@ -71,9 +71,9 @@ where
 /// tree.insert([0.2, 0.1]);
 /// tree.insert([0.3, 0.0]);
 ///
-/// assert_eq!(tree.nearest_neighbor(&[0.4, -0.1]), Some(&[0.3, 0.0]));
+/// assert_eq!(tree.nearest_neighbor([0.4, -0.1]), Some(&[0.3, 0.0]));
 /// tree.remove(&[0.3, 0.0]);
-/// assert_eq!(tree.nearest_neighbor(&[0.4, 0.3]), Some(&[0.2, 0.1]));
+/// assert_eq!(tree.nearest_neighbor([0.4, 0.3]), Some(&[0.2, 0.1]));
 ///
 /// assert_eq!(tree.size(), 2);
 /// // &RTree implements IntoIterator!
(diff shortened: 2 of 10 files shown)


rstar/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Added
 - Added missing re-exports of nearest neighbor iterators so they can be named in downstream crates. ([PR](https://github.com/georust/rstar/pull/186))
+- Made `RStar` methods take `Point` and `Envelope` as owned values where it makes sense ([PR](https://github.com/georust/rstar/pull/189))
```
adamreichold, line 5: I think this would rather belong into a `## Changed` section.

rstar/src/rtree.rs
```diff
@@ -327,51 +327,48 @@ where
...
     /// let unit_square = AABB::from_corners([0.0, 0.0], [1.0, 1.0]);
-    /// let elements_in_half_unit_square = tree.locate_in_envelope(&half_unit_square);
-    /// let elements_in_unit_square = tree.locate_in_envelope(&unit_square);
+    /// let elements_in_half_unit_square = tree.locate_in_envelope(half_unit_square);
+    /// let elements_in_unit_square = tree.locate_in_envelope(unit_square);
     /// assert_eq!(elements_in_half_unit_square.count(), 2);
     /// assert_eq!(elements_in_unit_square.count(), 3);
     /// ```
-    pub fn locate_in_envelope(&self, envelope: &T::Envelope) -> LocateInEnvelope<T> {
-        LocateInEnvelope::new(&self.root, SelectInEnvelopeFunction::new(envelope.clone()))
```
adamreichold, line 336: So removing the `.clone()` call is definitely an improvement as the combination of taking a reference and cloning is inflexible. I do wonder however if we should not go the other way around and make `LocateInEnvelope` actually take a reference to the `envelope`? It already borrows from `&self` so there should not a relevant loss of functionality if it borrows from `envelope` as well?

kristoff3r: I don't think that's a good idea for 2 reasons:
* The lifetimes aren't the same, as the envelope is not stored as part of the tree. My use case was to have a `LocateInEnvelope` that only depended on the lifetime of the tree.
* It looks to me like the envelope is almost always a tiny struct, so from a performance perspective it should be better to just have it owned by `SelectInEnvelopeFunction` and not go through an extra reference.

It would also be a much more invasive change I think, this only changes the outer API to reflect what the internals are already doing.

adamreichold: Agreed, the envelope will usually be a temporary on the stack and artificially extending its lifetime to keep the iterator viable is probably an unnecessary hassle due to the second observation.

Issue #84, opened 2022-01-20
Unique/hashable ID for each RTreeNode


giorgostheo:
From what I have seen, the current RTreeNode implementation does not offer a unique (hashable, implements == correctly) key for each RTreeNode.

More specifically, I believe there are many usecases where one might need some sort of internal "access" to the node, meaning a way to add some sort of data to each one. In my case, this "data" is a tuple of values. However, since there is nothing correctly hashable in each RTreeNode, I can not find a way to create a Hashmap for ex. in order to store the data I need. The only info that is always available for each RTreeNode is the envelope, however two nodes can have the same envelope which means that this is not an appropriate key. 

Adding some way to extract a (preferably hashable) unique ID for each RTreeNode would be useful in many ways, since it can allow  a more "free" approach to traversal and can also allow child to parent referances instead of just parent to child.

urschrei:
> a way to add some sort of data to each one

Does https://docs.rs/rstar/latest/rstar/primitives/struct.GeomWithData.html work for this?

giorgostheo:
No, the data needs to be added to each RTreeNode not to each individual point.

danieleades:
> > I have a use case where I need to prune whole subtrees before descending during a search, based on additional values stored in each internal node (aggregates summarising the leaves beneath it).
> > To support this, rstar would need to maintain custom per-node aggregates and update them automatically when the tree is rebalanced or items are inserted or removed. This implies:
> > 
> > 1. The ability to define what data is stored at each node
> > 2. Automatic recomputation of this data during splits, reinserts, and deletions
> > 3. A query API that can access both the envelope and the aggregate before deciding whether to descend
> 
> It is unlikely that anyone will add this feature for you; it's complex, somewhat niche, and would involve non-trivial changes to the tree internals if it proved possible to make those changes in a way that was zero-cost in the "you only pay for what you use" sense. I would encourage you to explore the codebase and determine whether it's feasible (feel free to open an issue / discussion for specific questions) – if you build up some promising momentum on the feature it's entirely possible that someone will help you to move it forward.

Totally agree this is too big a change to add directly.

What might be more realistic is adding hooks so that aggregates could be tracked externally rather than stored in the tree itself. That would require:

A way to uniquely reference nodes in the tree

Lifecycle hooks so aggregates can be updated when the structure changes (inserts, splits, deletions)

A traversal API that visits nodes in top-down order and exposes their envelopes and IDs

[...]

PR #200, opened 2025-08-14
Add tree traversal documentation and node counting examples

urschrei:
---

[See #84 ](https://github.com/georust/rstar/issues/84#issuecomment-3185546801) for rationale.


adamreichold (approved):
It might be nicer to use non-commutative operation like building up a string to highlight the difference via the differing results, but this certainly works.

urschrei:
Oh, good idea. Let me have a look.

PR #201, opened 2025-08-15
Fix Clippy warning (surfaced in Rust 1.89) related to lifetime elision

urschrei:
---

### DIFF
--- a/rstar/src/rtree.rs
+++ b/rstar/src/rtree.rs
@@ -294,7 +294,7 @@ where
     ///     println!("This tree contains point {:?}", point);
     /// }
     /// ```
-    pub fn iter(&self) -> RTreeIterator<T> {
+    pub fn iter(&self) -> RTreeIterator<'_, T> {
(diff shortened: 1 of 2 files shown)


rstar/src/rtree.rs
```diff
@@ -294,7 +294,7 @@ where
     ///     println!("This tree contains point {:?}", point);
     /// }
     /// ```
-    pub fn iter(&self) -> RTreeIterator<T> {
+    pub fn iter(&'_ self) -> RTreeIterator<'_, T> {
```
adamreichold, line 297:
```suggestion
    pub fn iter(&self) -> RTreeIterator<'_, T> {
```

I think the left-hand side is generally unnecessary here to silence the lint (since `&` implies a lifetime) and I would also prefer it.

adamreichold: Nice. Could you apply it uniformly to all (self) references?

adamreichold (approved):

PR #204, opened 2025-08-16
Simplify CI

urschrei:
We no longer use `cargo-all-features` in geo, and rstar doesn't currently have any features, so this strategy needlessly inflates our test times.

Also bumps Docker images to test on Rust 1.83 and 1.84, retaining 1.63

---


urschrei:
I've also added caching to try to speed things up. Notably, the 1.63 job currently takes 6 minutes (vs around 1 minute for the 1.8x job), presumably due to the older index crates.io sync mechanism

.github/workflows/test.yml
```diff
@@ -35,17 +35,20 @@ jobs:
           # Minimum supported rust version (MSRV)
           - "georust/geo-ci:rust-1.63"
```
adamreichold, line 36: This was never pushed to ghcr.io?

.github/workflows/test.yml
```diff
@@ -35,17 +35,20 @@ jobs:
...
-      - run: cargo build-all-features
-      - run: cargo test-all-features
-      - run: cargo build -p rstar-benches
+      - name: Setup Rust cache
+        uses: Swatinem/rust-cache@v2
+        with:
+          workspaces: "rstar -> target"
+          key: ${{ matrix.container_image }}
+      - run: cargo check --all-targets
+      - run: cargo test
```
adamreichold, line 51: Shouldn't these have a `--workspace` added?

Also, the `cargo build -p rstar-benches` command does not seem to covered, is this intentional?

.github/workflows/test.yml
```diff
@@ -57,6 +60,10 @@ jobs:
         uses: dtolnay/rust-toolchain@stable
         with:
           components: clippy, rustfmt
+      - name: Setup Rust cache
+        uses: Swatinem/rust-cache@v2
+        with:
+          workspaces: "rstar -> target"
```
adamreichold, line 66: Our workspace is in the top-level folder and the `rstar` contains the package by the same name, so I upon reading https://github.com/Swatinem/rust-cache?tab=readme-ov-file#example-usage, I would expect the default of `. -> target` to fit our use case?

adamreichold: What I meant to say is that you can remove the `with: workspaces: ..` part everywhere. The defaults should work fine.

urschrei:
Ah, the `workspace` directive is causing the build to break because `geo` is using the old API which borrowed geometries (changed in https://github.com/georust/rstar/pull/189).

I'll have to fix this on the geo side and build new images before we can merge this.

urschrei:
Seems sensible, and runs are all completing with the new setup.

.github/workflows/test.yml
```diff
@@ -33,34 +32,53 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v5
+      - name: Install stable toolchain
+        uses: dtolnay/rust-toolchain@stable
+      - name: Setup Rust cache
+        uses: Swatinem/rust-cache@v2
+      - name: Build bench
+        run: cargo build --workspace --package rstar-benches
```
adamreichold, line 62: Are pure build tests really useful when we do Clippy builds for the whole workspace anyway? The main point here would be following through with codegen and linking?

.github/workflows/test.yml
```diff
@@ -33,34 +32,53 @@ jobs:
...
-      - run: cargo install --version 1.6.0 cargo-all-features
-      - run: cargo build-all-features
-      - run: cargo test-all-features
-      - run: cargo build -p rstar-benches
+        uses: actions/checkout@v5
+      - name: Setup Rust cache
+        uses: Swatinem/rust-cache@v2
+        with:
+          key: ${{ matrix.container_image }}
+      - run: cargo build --package rstar
```
adamreichold, line 48: Do we really need to build if we do a test build afterwards anyway?

.github/workflows/test.yml
urschrei: Removed the build job for `bench` and `demo` entirely since we do nothing with the result, as you say, and the Clippy job covers the build.

adamreichold (approved):

PR #211, opened 2025-09-22
Update rtree object: impl RTreeObject and PointDistance for Arc<T>

QIU1995NONAME:
---


QIU1995NONAME:
Hi, I originally wanted to implement something like this:
```rust
impl<P, T> RTreeObject for P
where
    P: Deref<Target = T> + Drop,
    T: RTreeObject + ?Sized,
```

or

```rust
impl<P, T> RTreeObject for P
where
    P: Borrow<T> + Drop,
    T: RTreeObject + ?Sized,
```

However, it conflicts with the existing implementations.
If I just copy the existing ```impl<T> RTreeObject for Arc<T>``` block, it ends up duplicating quite a bit of code.

Do you have any suggestions on how to handle this more cleanly?

Thanks!

PR #223, opened 2026-03-02
Avoid unclear diameter terminology in AABB::from_center.

adamreichold:
---

### DIFF
--- a/rstar/src/aabb.rs
+++ b/rstar/src/aabb.rs
@@ -71,14 +71,14 @@ where
         Self { lower, upper }
     }
 
-    /// Creates a new AABB from a center and a diameter.
-    pub fn from_center(center: P, diameter: P::Scalar) -> Self {
-        let one = P::Scalar::one();
-        let two = one + one;
-        let radius = P::from_value(diameter / two);
+    /// Creates a new AABB from a center and a distance.
+    ///
+    /// Creates the smallest AABB which includes all points within `distance` of `center`.
+    pub fn from_center(center: P, distance: P::Scalar) -> Self {
+        let distance = P::from_value(distance);
 
-        let p1 = center.add(&radius);
-        let p2 = center.sub(&radius);
+        let p1 = center.add(&distance);
+        let p2 = center.sub(&distance);
 
         Self::from_corners(p1, p2)
     }


adamreichold:
Trying to address https://github.com/georust/rstar/pull/222#discussion_r2862898905

michaelkirk (commented):
I was expecting "width". I guess it's a little funny because width == height. 

To my (US_en) eye, "distance" is more or less synonymous with diameter.

In particular: the distance from the center to the edge is not the same as the distance from the center to the corner. 

Is there another popular rectangle API that uses this  "distance" term? Everything I find is either min/max points or origin,width,height.

I can't find another implementation of our funny width==height constructor.

adamreichold:
My idea behind the choice of "distance" was that this is the smallest AABB which includes all points within the given distance of the center? (This should even be true no matter which metric on R^n is chosen, shouldn't it?)

However, I am personally not particularly invested in any terminology and will happily change this if you have a clear favourite? (Or anybody else for that matter.)

> To my (US_en) eye, "distance" is more or less synonymous with diameter.

At least to me, "distance" would be synonymous to radius instead of diameter.

adamreichold:
@urschrei @michaelkirk I think with this resolved, a new release would be a good thing(tm). Of course, we need to resolve the terminology discussion here in some way. I am not particularly invested in any specific wording and would happily apply changes if one of you has a strong opinion. Until then, I think "distance" as proposed here is the least bad option so far, especially considering

> this is the smallest AABB which includes all points within the given distance of the center

So maybe I add this to the doc string and we run with this?

urschrei:
> this is the smallest AABB which includes all points within the given distance of the center

I think this is unambiguous. We're long overdue for a new release of rstar – @michaelkirk  WDYT?

urschrei (approved):

michaelkirk:
Sorry to have held this up.

To be clear, this is not only a terminology change, but a change in behavior, right? Previous callers will get back an AABB that is twice the width as before. Right? It's not yet released, so I think that's OK. Hopefully folks aren't using the pre-release version.

Feel free to merge with what you have if you're happy with it.

urschrei:
> Hopefully folks aren't using the pre-release version.

If you're doing that and not pinning to a commit you're asking for trouble, though.

0.13.0 will be a breaking release due to the ownership change in https://github.com/georust/rstar/pull/189. We should prominently mark that in the changelog.

PR #227, opened 2026-03-10
Rtrees in one dimension

leo-collins:
---

Allow construction of 1D rtrees. I can't see at a glance why this shouldn't be okay. Added some basic tests.


adamreichold:
We discussed this in https://github.com/georust/rstar/pull/169 and decided against it. Could you check that discussion if something has significantly changed since then? Thanks.

leo-collins:
Basically allowing this would save us having to write a separate tree data structure and special-case it in several places for 1D. In our use case we're definitely not going to run into the situation that this would be slow in 1D.

Perhaps a note in the docs saying "1D is supported but is potentially a stupid idea"? If not then I will close this PR.

PR #228, opened 2026-03-27
create an AABB from known bounds

mockersf:
---

I tried using rstar, and had significantly worse results than I expected. In my benchmarks, I noticed the code was spending a lot of times finding min_point/max_point for the AABB construction even though I already knew them.

This PR adds a new utility constructor that takes known bounds, and asserts they are correct in debug mode

### DIFF
--- a/rstar/src/aabb.rs
+++ b/rstar/src/aabb.rs
@@ -64,6 +64,13 @@ where
         }
     }
 
+    /// Returns the AABB from already known lower/upper bounds.
+    pub fn from_bounds(lower: P, upper: P) -> Self {
+        debug_assert!(lower.min_point(&upper) == lower);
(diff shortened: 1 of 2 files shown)


rstar/src/aabb.rs
```diff
@@ -109,6 +109,13 @@ where
             self.min_point(point).sub(point).length_2()
         }
     }
+
+    /// Returns the AABB from already known lower/upper bounds.
+    pub fn from_bounds(lower: P, upper: P) -> Self {
```
adamreichold: Could you please move this up next to `from_corners`? Also please unify usage of `Self` instead of `AABB` in these two methods while we are here.

mockersf: done!

adamreichold (approved):

PR #229, opened 2026-05-14
Modernize dependencies, most importantly heapless to avoid unmaintained atomic-polyfill.

adamreichold:

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,37 +18,22 @@ env:
   CARGO_TERM_COLOR: always
 
 jobs:
-  rstar:
-    name: rstar
+  test:
+    name: Unit tests
     runs-on: ubuntu-latest
-    if: "!contains(github.event.head_commit.message, '[skip ci]')"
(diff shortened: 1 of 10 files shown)


urschrei (approved):
Not clear why that run is failing, but happy for this to merge.

adamreichold:
> Not clear why that run is failing, but happy for this to merge.

rand 0.10 has an MSRV of 1.85 but it is only a dev dependency so I am trying to figure out if I can simplify our CI to make MSRV testing more targetted so it really only builds the library.

adamreichold:
Ok, so I don't think MSRV 1.68 is worth fuzzing over and following `rand` to a 1.85 MSRV is reasonable as this is currently what Debian stable (a.k.a. Trixie) ships. (Still fighting with heapless on its 1.87 MSRV though.)

urschrei:
> 1.85 MSRV is reasonable as this is currently what Debian stable (a.k.a. Trixie) ships

Yep, totally fine.

PR #232, opened 2026-06-04
benches: drop geo-types dependency, fix rand 0.10 API

urschrei:
The 0.13 release left rstar-benches uncompilable: rand 0.10 moved the random-value methods from Rng to RngExt (gen() -> random()), and geo-types 0.7.19 only binds rstar <= 0.12, so the workspace rstar patch no longer applied and the Polygon benches lost their RTreeObject impl.

Replace geo_types::Polygon with a local BenchPolygon (RTreeObject whose envelope is the AABB of its ring), dropping the geo-types dependency, and switch to RngExt::random().

---

### DIFF
--- a/rstar-benches/benches/benchmarks.rs
+++ b/rstar-benches/benches/benchmarks.rs
@@ -1,18 +1,16 @@
 #[macro_use]
 extern crate criterion;
-extern crate geo_types;
 extern crate rand;
 extern crate rand_hc;
 extern crate rstar;
 
 use std::f64::consts::PI;
 
-use geo_types::{LineString, Polygon};
-use rand::{Rng, SeedableRng};
+use rand::{RngExt, SeedableRng};
 use rand_hc::Hc128Rng;
 
 use rstar::primitives::CachedEnvelope;
-use rstar::{RStarInsertionStrategy, RTree, RTreeParams};
+use rstar::{RStarInsertionStrategy, RTree, RTreeObject, RTreeParams, AABB};
 
 use criterion::Criterion;
 
@@ -58,7 +56,7 @@ fn bulk_load_complex_geom(c: &mut Criterion) {
             create_random_polygons(DEFAULT_BENCHMARK_TREE_SIZE, 4096, SEED_1).collect();
 
         b.iter(|| {
-            RTree::<Polygon<f64>, Params>::bulk_load_with_params(polys.clone());
+            RTree::<BenchPolygon, Params>::bulk_load_with_params(polys.clone());
         });
     });
 }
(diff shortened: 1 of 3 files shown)


adamreichold (approved):
I admit to having explicitly made the CI not depend on the benchmarks, but this is mostly because I think this is a fundamental issues. `geo-types` can always support only released versions of `rstar` and so we would always have to accept internal breakage once we publish a breaking release. Hence, I am glad about this changing breaking the circular dependency.

Taking a bird's eye perspective, as mentioned in previous discussions on that topic, I still think that we should consider `geo-types` as a "vocabulary crate" similar to `serde` have everybody else including `rstar` depend on it instead of the other way around.

adamreichold:
> I admit to having explicitly made the CI not depend on the benchmarks, but this is mostly because I think this is a fundamental issues.

Of course, we could consider including it back in the CI now that builds independently of `geo-types`?

urschrei:
> I still think that we should consider geo-types as a "vocabulary crate" similar to serde have everybody else including rstar depend on it instead of the other way around.

This is interesting. One issue we may have there is that we've had ongoing discussions about the possibility of merging `geo-types` back into geo, as the split is causing issues with e.g. rstar. I confess that I don't know what the correct approach is, and we're certainly nowhere near a decision.

> Of course, we could consider including it back in the CI now that builds independently of geo-types?

I would be in favour of doing so, I wonder whether it would be possible to highlight perf regressions (assuming we don't want to fail a build outright)

adamreichold:
> I would be in favour of doing so, I wonder whether it would be possible to highlight perf regressions (assuming we don't want to fail a build outright)

I think this is difficult to do using plain GitHub Actions, but there are vendors like [CodSpeed](https://codspeed.io/) which specialize in that kind of tracking and currently free-of-charge for FOSS projects IIRC. You could create an account there and set up the integration which would include running the benchmarks using their harness.

adamreichold:
> One issue we may have there is that we've had ongoing discussions about the possibility of merging geo-types back into geo, as the split is causing issues with e.g. rstar. 

My instinct here is that `geo-types` should really only contain data structures, i.e. vocabulary on which everybody working geographic information has to agree on, and should not be merged into `geo` which in my mind is mostly about algorithms to solve common problems. `rstar` would then work to support as many `geo-types` versions as reasonably possible and `geo` would just need to use an `rstar` release compatible with its `geo-types` version. Ideally, doing another release of `rstar` to support a new version of `geo-types` should be simple and non-breaking (but also happen rarely as the vocabulary should not change that often).

michaelkirk (approved):

rstar-benches/benches/benchmarks.rs
```diff
@@ -156,42 +154,57 @@ criterion_main!(benches);
...
-    Polygon::new(LineString(ring), Vec::new())
+    BenchPolygon { ring }
+}
+
+/// A minimal stand-in for a `geo_types::Polygon` used only to exercise bulk loading
+/// of a non-trivial [`RTreeObject`] whose envelope spans many vertices. Its envelope
+/// is the axis-aligned bounding box of the exterior ring; no point queries are run
+/// against it, so it deliberately implements only [`RTreeObject`].
+#[derive(Clone)]
+struct BenchPolygon {
```
michaelkirk, line 200: I don't think you even need to mention geo-types. You can just call this `Polygon`. The historical reliance on geo-types isn't particularly interesting in my opinion.

But this really doesn't matter, so take it or leave it.

PR #236, opened 2026-07-28
feat(rtree): Add support for `AsRef`/`AsMut` for `RTree<_, _>`

fogti:
The purpose of this is to properly get rid of
https://docs.rs/rstared/0.13.1/rstared/struct.AsRefRTree.html, which only exists to introduce such an `AsRef<RTree<_, _>>` implementation.

---

### DIFF
--- a/rstar/src/rtree.rs
+++ b/rstar/src/rtree.rs
@@ -219,6 +219,28 @@ where
     }
 }
 
+impl<T, Params> AsRef<RTree<T, Params>> for RTree<T, Params>
+where
+    T: RTreeObject,
+    Params: RTreeParams,
+{
+    #[inline]
+    fn as_ref(&self) -> &RTree<T, Params> {
+        self
+    }
+}
+
+impl<T, Params> AsMut<RTree<T, Params>> for RTree<T, Params>
+where
+    T: RTreeObject,
+    Params: RTreeParams,
+{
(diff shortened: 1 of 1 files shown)


adamreichold (approved):
These are weird, but apparently this can be necessary in generic code and there is precedent in the standard library, e.g. <https://doc.rust-lang.org/stable/std/convert/trait.AsRef.html#impl-AsRef%3Cstr%3E-for-str> or <https://doc.rust-lang.org/stable/std/convert/trait.AsMut.html#impl-AsMut%3C%5BT%5D%3E-for-%5BT%5D>.

EDIT: I wonder where there is no blanket impl for the identity case...

fogti:
> I wonder where there is no blanket impl for the identity case...

afaik simply because the rust type systems doesn't have sufficient features to make sure some other generic impls can never overlap with that, particularly when the impl wants to skip over some indirection...

adamreichold:
Merged. Thank you for your contribution!

fogti:
Would it be possible to get a release of `rstar` with this included? It would unblock some downstream work (albeit mostly in regards to decoupling of and probably deprecating the current state of `rstared`).

urschrei:
We just need to finish https://github.com/georust/rstar/pull/237 because it's a little more impactful than it first appeared.

PR #237, opened 2026-08-11
Don't panic on non-finite bounds

### LINKED ISSUE #112: Panic after inserting `Rectangle` with NaN to `RTree`
I wrote a program that generates random shapes and inserts them into a `RTree`. I accidentally generated some shapes that contained NaN and to my surprise, the program panicked. The panic occurred inside rstar.

Here a minimal example that I've tested with rstar 0.10.0
```rust
use rstar::{RTree, primitives::Rectangle};

fn main() {
    let mut tree = RTree::new();
    tree.insert(Rectangle::from_corners([-1.4, 1.0], [-1.3, 1.1]));
    tree.insert(Rectangle::from_corners([2.3, 0.0], [2.4, 0.0]));
    tree.insert(Rectangle::from_corners([-1.4, -0.4], [-1.4, -0.2]));
    tree.insert(Rectangle::from_corners([f32::NAN, f32::NAN], [f32::NAN, f32::NAN]));
    tree.insert(Rectangle::from_corners([1.8, -1.8], [1.9, -1.8]));
    tree.insert(Rectangle::from_corners([0.2, -0.5], [0.3, -0.4]));
    tree.insert(Rectangle::from_corners([-0.6, -0.6], [-0.4, -0.5])); // This will panic!
}
```
Output:
```text
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', /home/explorer/.cargo/registry/src/github.com-1ecc6299db9ec823/rstar-0.10.0/src/algorithm/rstar.rs:345:14
```
[Rust explorer link](https://www.rustexplorer.com/b#%2F*%0A%5Bdependencies%5D%0Arstar%20%3D%20%220.10.0%22%0A*%2F%0A%0Ause%20rstar%3A%3A%7BRTree%2C%20primitives%3A%3ARectangle%7D%3B%0A%0Afn%20main()%20%7B%0A%20%20%20%20let%20mut%20tree%20%3D%20RTree%3A%3Anew()%3B%0A%20%20%20%20tree.insert(Rectangle%3A%3Afrom_corners(%5B-1.4%2C%201.0%5D%2C%20%5B-1.3%2C%201.1%5D))%3B%0A%20%20%20%20tree.insert(Rectangle%3A%3Afrom_corners(%5B2.3%2C%200.0%5D%2C%20%5B2.4%2C%200.0%5D))%3B%0A%20%20%20%20tree.insert(Rectangle%3A%3Afrom_corners(%5B-1.4%2C%20-0.4%5D%2C%20%5B-1.4%2C%20-0.2%5D))%3B%0A%20%20%20%20tree.insert(Rectangle%3A%3Afrom_corners(%5Bf32%3A%3ANAN%2C%20f32%3A%3ANAN%5D%2C%20%5Bf32%3A%3ANAN%2C%20f32%3A%3ANAN%5D))%3B%0A%20%20%20%20tree.insert(Rectangle%3A%3Afrom_corners(%5B1.8%2C%20-1.8%5D%2C%20%5B1.9%2C%20-1.8%5D))%3B%0A%20%20%20%20tree.insert(Rectangle%3A%3Afrom_corners(%5B0.2%2C%20-0.5%5D%2C%20%5B0.3%2C%20-0.4%5D))%3B%0A%20%20%20%20tree.insert(Rectangle%3A%3Afrom_corners(%5B-0.6%2C%20-0.6%5D%2C%20%5B-0.4%2C%20-0.5%5D))%3B%20%2F%2F%20This%20will%20panic!%0A%7D)

Is this intentional behavior?

michaelkirk:
---

Fixes https://github.com/georust/rstar/issues/112

Our old friend, NaN. I noticed this while trying to use the [new (unreleased) validation feature](https://github.com/georust/geo/pull/1522) in geo which can panic when attempting to validate Polygons with NaN in their interior.

This is one of those cases where we swap a `--release` panic for potentially giving incorrect / unexpected results. Is it a good idea? Reasonable minds may disagree! But panic'ing in `release` mode is a pretty serious offense.

### DIFF
--- a/rstar/Cargo.toml
+++ b/rstar/Cargo.toml
@@ -21,6 +21,7 @@ num-traits = { version = "0.2", default-features = false, features = ["libm"] }
 serde = { version = "1.0", optional = true, default-features = false, features = ["alloc", "derive"] }
 smallvec = "1.6"
 mint = { version = "0.5.9", optional = true }
+ordered-float = { workspace = true }
 
 [features]
 default = []
--- a/rstar/Cargo.toml
+++ b/rstar/Cargo.toml
@@ -8,7 +8,7 @@ repository = "https://github.com/georust/rstar"
 license = "MIT OR Apache-2.0"
 readme = "README.md"
 edition = "2021"
-rust-version = "1.85"
+rust-version = "1.90"
 keywords = ["rtree", "r-tree", "spatial", "spatial-index", "nearest-neighbor"]
 categories = ["data-structures", "algorithms", "science::geo"]
 
--- a/rstar/src/aabb.rs
+++ b/rstar/src/aabb.rs
@@ -1,4 +1,4 @@
-use crate::point::{max_inline, Point, PointExt};
+use crate::point::{max_inline, Point, PointExt, RTreeNum};
 use crate::{Envelope, RTreeObject};
 use num_traits::{Bounded, One, Zero};
 
@@ -134,17 +134,23 @@ where
     }
 
     fn is_empty(&self) -> bool {
-        self.lower.nth(0) > self.upper.nth(0)
+        self.lower.nth(0).ord() > self.upper.nth(0).ord()
     }
 
     fn contains_point(&self, point: &P) -> bool {
-        self.lower.all_component_wise(point, |x, y| x <= y)
-            && self.upper.all_component_wise(point, |x, y| x >= y)
+        self.lower
+            .all_component_wise(point, |x, y| x.ord() <= y.ord())
+            && self
+                .upper
+                .all_component_wise(point, |x, y| x.ord() >= y.ord())
     }
 
     fn contains_envelope(&self, other: &Self) -> bool {
-        self.lower.all_component_wise(&other.lower, |l, r| l <= r)
-            && self.upper.all_component_wise(&other.upper, |l, r| l >= r)
+        self.lower
+            .all_component_wise(&other.lower, |l, r| l.ord() <= r.ord())
+            && self
+                .upper
+                .all_component_wise(&other.upper, |l, r| l.ord() >= r.ord())
     }
 
     fn merge(&mut self, other: &Self) {
@@ -160,8 +166,11 @@ where
     }
 
     fn intersects(&self, other: &Self) -> bool {
-        self.lower.all_component_wise(&other.upper, |l, r| l <= r)
-            && self.upper.all_component_wise(&other.lower, |l, r| l >= r)
+        self.lower
+            .all_component_wise(&other.upper, |l, r| l.ord() <= r.ord())
+            && self
+                .upper
+                .all_component_wise(&other.lower, |l, r| l.ord() >= r.ord())
     }
 
     fn area(&self) -> P::Scalar {
@@ -178,7 +187,8 @@ where
     fn min_max_dist_2(&self, point: &P) -> <P as Point>::Scalar {
         let l = self.lower.sub(point);
         let u = self.upper.sub(point);
-        let mut max_diff = (Zero::zero(), Zero::zero(), 0); // diff, min, index
+        // diff, min, index
+        let mut max_diff: (P::Scalar, P::Scalar, usize) = (Zero::zero(), Zero::zero(), 0);
         let mut result = P::new();
 
         for i in 0..P::DIMENSIONS {
@@ -186,14 +196,14 @@ where
             let mut max = u.nth(i);
             max = max * max;
             min = min * min;
-            if max < min {
+            if max.ord() < min.ord() {
                 core::mem::swap(&mut min, &mut max);
             }
 
             let diff = max - min;
             *result.nth_mut(i) = max;
 
-            if diff >= max_diff.0 {
+            if diff.ord() >= max_diff.0.ord() {
                 max_diff = (diff, min, i);
             }
         }
@@ -227,8 +237,8 @@ where
             l.envelope()
                 .lower
                 .nth(axis)
-                .partial_cmp(&r.envelope().lower.nth(axis))
-                .unwrap()
+                .ord()
+                .cmp(&r.envelope().lower.nth(axis).ord())
         });
     }
 
@@ -241,8 +251,8 @@ where
             l.envelope()
                 .lower
                 .nth(axis)
-                .partial_cmp(&r.envelope().lower.nth(axis))
-                .unwrap()
+                .ord()
+                .cmp(&r.envelope().lower.nth(axis).ord())
         });
     }
 }
--- a/rstar/src/aabb.rs
+++ b/rstar/src/aabb.rs
@@ -378,4 +378,91 @@ mod test {
         tree.insert([-0.0f64, -0.0]);
         assert_eq!(tree.locate_at_point([0.0f64, 0.0]), Some(&[-0.0, -0.0]));
     }
+
+    #[test]
+    fn not_nan_coordinates_are_supported() {
+        use ordered_float::NotNan;
+
+        let point = |x: f64, y: f64| [NotNan::new(x).unwrap(), NotNan::new(y).unwrap()];
+
+        let mut tree = RTree::new();
+        for value in 0..32 {
+            tree.insert(point(value as f64, (value * 2) as f64));
+        }
+
+        assert_eq!(tree.size(), 32);
+        assert_eq!(
+            tree.locate_at_point(point(10.0, 20.0)),
+            Some(&point(10.0, 20.0))
+        );
+        assert_eq!(
+            tree.nearest_neighbor(point(10.4, 20.0)),
+            Some(&point(10.0, 20.0))
+        );
+        assert_eq!(tree.locate_at_point(point(0.5, 0.5)), None);
+    }
+
+    // document some known panic cases with NotNan
+    mod known_panics {
+        use super::*;
+
+        #[test]
+        #[should_panic(expected = "resulted in NaN")]
+        fn infinite_not_nan_coordinates_panic() {
+            use ordered_float::NotNan;
+
+            let point = |x: f64, y: f64| [NotNan::new(x).unwrap(), NotNan::new(y).unwrap()];
+            let infinite = point(f64::INFINITY, 0.0);
+
+            let mut tree = RTree::new();
+            tree.insert(infinite);
+            tree.insert(point(0.0, 0.0));
+
+            // This line panics. Measuring the distance from the query point to the point at infinity computes
+            // `inf - inf`, which is `NaN`.
+            tree.nearest_neighbor(infinite);
+        }
+
+        /// `area` multiplies the envelope's
+        /// extents together, once per dimension, so in 3-D any extent beyond the cube root of
+        /// `f32::MAX` (7e12) gives `inf`. `choose_subtree` subtracts two such areas, and `inf - inf`
+        /// is `NaN` -- though every coordinate below is finite and well inside `f32::MAX` (3.4e38).
+        #[test]
+        #[should_panic(expected = "resulted in NaN")]
+        fn overflowing_not_nan_coordinates_panic() {
+            use ordered_float::NotNan;
+
+            let point = |i: u32| {
+                let v = 1e13 * i as f32;
+                [
+                    NotNan::new(v).unwrap(),
+                    NotNan::new(-v).unwrap(),
+                    NotNan::new(v / 2.0).unwrap(),
+                ]
+            };
+
+            let mut tree = RTree::new();
+            for i in 0..8 {
+                tree.insert(point(i));
+            }
+        }
+
+        /// The likelier failure in practice needs no extreme magnitudes at all: `nearest_point`
+        /// divides by the line's squared length, so a segment whose endpoints coincide -- a polyline
+        /// with a duplicated vertex -- divides `0.0 / 0.0` while answering a query.
+        #[test]
+        #[should_panic(expected = "resulted in NaN")]
+        fn degenerate_not_nan_line_panics() {
+            use crate::primitives::Line;
+            use ordered_float::NotNan;
+
+            let n = |v: f64| NotNan::new(v).unwrap();
+
+            let mut tree = RTree::new();
+            tree.insert(Line::new([n(0.0), n(0.0)], [n(1.0), n(1.0)]));
+            tree.insert(Line::new([n(2.0), n(2.0)], [n(2.0), n(2.0)]));
+
+            tree.nearest_neighbor([n(2.0), n(3.0)]);
+        }
+    }
 }
--- a/rstar/src/aabb.rs
+++ b/rstar/src/aabb.rs
@@ -252,6 +252,7 @@ mod test {
     use super::AABB;
     use crate::envelope::Envelope;
     use crate::object::PointDistance;
+    use crate::RTree;
 
     #[test]
     fn empty_rect() {
@@ -300,5 +301,71 @@ mod test {
 
         let not_empty = AABB::from_corners([1.0, 1.0], [1.0, 1.0]);
         assert!(!not_empty.is_empty());
+    }
+
+    #[test]
+    fn rtree_operations_with_nan_do_not_panic() {
+        let mut points: Vec<_> = (0..64).map(|value| [value as f64, value as f64]).collect();
+        points[16][0] = f64::NAN;
+
+        let tree = RTree::bulk_load(points.clone());
+        assert_eq!(tree.size(), points.len());
+        assert_eq!(
+            tree.nearest_neighbor_iter([f64::NAN, 0.0]).count(),
+            points.len()
+        );
+
+        let mut tree = RTree::new();
+        for point in &points {
+            tree.insert(*point);
+        }
+
+        assert_eq!(tree.size(), points.len());
+        assert_eq!(
+            tree.nearest_neighbor_iter([f64::NAN, 0.0]).count(),
+            points.len()
+        );
+    }
+
+    #[test]
+    fn tree_with_nan_is_still_queryable() {
+        let before_the_nan = [10.0, 0.0];
+        let after_the_nan = [20.0, 0.0];
+
+        let mut tree = RTree::new();
+        tree.insert(before_the_nan);
+        tree.insert([f64::NAN, 0.0]);
+        tree.insert(after_the_nan);
+
+        assert_eq!(tree.size(), 3);
+        assert_eq!(tree.iter().count(), 3);
+
+        assert_eq!(tree.locate_at_point(before_the_nan), Some(&before_the_nan));
+        assert_eq!(tree.locate_at_point(after_the_nan), Some(&after_the_nan));
+
+        assert!(tree.contains(&before_the_nan));
+        assert!(tree.contains(&after_the_nan));
+
+        assert_eq!(tree.remove(&before_the_nan), Some(before_the_nan));
+        assert_eq!(tree.remove(&after_the_nan), Some(after_the_nan));
+    }
+
+    /// `-0.0` and `0.0` are the same point as far as `==` is concerned, so a query for
+    /// one must find data inserted under the other.
+    ///
+    /// This is why the scalar ordering is `OrderedFloat` rather than `f64::total_cmp`:
+    /// which would rank `-0.0` strictly below `0.0`.
+    #[test]
+    fn signed_zeroes_are_interchangeable_in_queries() {
+        let envelope = AABB::from_corners([0.0f64, 0.0], [1.0, 1.0]);
+        assert!(envelope.contains_point(&[-0.0f64, -0.0]));
+
+        let mut tree = RTree::new();
+        tree.insert([0.0f64, 0.0]);
+        assert_eq!(tree.locate_at_point([-0.0f64, -0.0]), Some(&[0.0, 0.0]));
+
+        let mut tree = RTree::new();
+        tree.insert([-0.0f64, -0.0]);
+        assert_eq!(tree.locate_at_point([0.0f64, 0.0]), Some(&[-0.0, -0.0]));
     }
 }
--- a/rstar/src/algorithm/nearest_neighbor.rs
+++ b/rstar/src/algorithm/nearest_neighbor.rs
@@ -1,4 +1,4 @@
-use crate::point::min_inline;
+use crate::point::{min_inline, RTreeNum};
 use crate::{
     node::{ParentNode, RTreeNode},
     object::Distance,
@@ -28,7 +28,9 @@ where
     T: PointDistance,
 {
     fn eq(&self, other: &Self) -> bool {
-        self.distance == other.distance
+        // Not `self.distance == other.distance`: that leaves `Eq` inconsistent with the
+        // `Ord` impl below, which reports two NaN distances as equal.
+        self.distance.ord() == other.distance.ord()
     }
 }
 
@@ -49,7 +51,7 @@ where
 {
     fn cmp(&self, other: &Self) -> ::core::cmp::Ordering {
         // Inverse comparison creates a min heap
-        other.distance.partial_cmp(&self.distance).unwrap()
+        other.distance.ord().cmp(&self.distance.ord())
     }
 }
 
@@ -248,7 +250,7 @@ where
             let distance_if_less_or_equal = match child {
                 RTreeNode::Parent(ref data) => {
                     let distance = data.envelope.distance_2(query_point);
-                    if distance <= *min_max_distance {
+                    if distance.ord() <= min_max_distance.ord() {
                         Some(distance)
                     } else {
                         None
--- a/rstar/src/point.rs
+++ b/rstar/src/point.rs
@@ -1,5 +1,7 @@
 use core::fmt::Debug;
+use core::num::Wrapping;
 use num_traits::{Bounded, Num, Signed, Zero};
+use ordered_float::OrderedFloat;
 
 /// Defines a number type that is compatible with rstar.
 ///
@@ -8,13 +10,15 @@ use num_traits::{Bounded, Num, Signed, Zero};
 ///  - [Wrapping](core::num::Wrapping) versions of the above
 ///  - f32, f64
 ///
-/// This type cannot be implemented directly. Instead, it is required to implement
-/// all required traits from the `num_traits` crate.
+/// r-trees require a *total* order on their scalars.
+/// [`RTreeNum::OrdType`] supplies that ordering. Integral types, which are already Ord,
+/// are their own OrdType. `f32` and `f64` use [`ordered_float::OrderedFloat`].
 ///
 /// # Example
 /// ```
 /// # extern crate num_traits;
 /// use num_traits::{Bounded, Num, Signed};
+/// use rstar::RTreeNum;
 ///
 /// #[derive(Clone, Copy, PartialEq, PartialOrd, Debug)]
 /// struct MyFancyNumberType(f32);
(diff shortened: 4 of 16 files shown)


rstar/src/aabb.rs
```diff
@@ -292,7 +285,6 @@ mod test {
         let aabb = AABB::from_points(&[(3., 3., 3.), (4., 4., 4.)]);
         assert_eq!(aabb, AABB::from_corners((3., 3., 3.), (4., 4., 4.)));
     }
-
```
adamreichold, line 295: Spurious whitespace change?

rstar/src/aabb.rs
```diff
@@ -301,4 +293,28 @@ mod test {
         let not_empty = AABB::from_corners([1.0, 1.0], [1.0, 1.0]);
         assert!(!not_empty.is_empty());
     }
+
+    #[test]
+    fn rtree_operations_with_nan_do_not_panic() {
```
adamreichold, line 314: Just to make sure: This fails before the switch to `total_cmp`?

rstar/src/lib.rs
```diff
@@ -9,6 +9,10 @@
 //! _Spatial data_ refers to an object that has the notion of a position and extent:
 //! for example points, lines and rectangles in any dimension.
 //!
+//! # NaN coordinates
+//! Geometries and query points must not contain `NaN` coordinates. Behavior of r-tree
+//! operations involving `NaN` coordinates is undefined.
```
adamreichold: I would suggest avoid the "undefined behaviour" terminology and rather suggest incorrect or arbitrary behaviour and explicitly state that we will not panic as a contract.

rstar/src/point.rs
```diff
@@ -99,6 +99,60 @@ pub trait RTreeNum: Bounded + Num + Clone + Copy + Signed + PartialOrd + Debug {
 
 impl<S> RTreeNum for S where S: Bounded + Num + Clone + Copy + Signed + PartialOrd + Debug {}
 
+/// Like `f64::total_cmp` but for all RTreeNum (including integers)
+pub(crate) fn total_cmp<S: RTreeNum>(left: S, right: S) -> core::cmp::Ordering {
+    match left.partial_cmp(&right) {
+        Some(ordering) => ordering,
+        // NaN is unordered even with itself, so sort NaN-like values after
+        // ordered values and group them together.
+        None => match (
```
adamreichold: Maybe we should move this out into a `#[cold]` function to avoid code bloat in the common case? Did you run any benchmarks against this change?

rstar/src/point.rs
```diff
@@ -99,6 +99,60 @@ pub trait RTreeNum: Bounded + Num + Clone + Copy + Signed + PartialOrd + Debug {
 
 impl<S> RTreeNum for S where S: Bounded + Num + Clone + Copy + Signed + PartialOrd + Debug {}
 
+/// Like `f64::total_cmp` but for all RTreeNum (including integers)
```
adamreichold: I wonder if we should funnel this through a `RTreeCmp` trait or something so that downstream code could override it and we can call `f32/f64::total_cmp`  directly?

rstar/src/aabb.rs
michaelkirk: Correct.

rstar/src/lib.rs
michaelkirk: Good point. "Undefined behavior" has a very specific meaning.

rstar/src/point.rs
michaelkirk: Benches are currently broken due to https://github.com/georust/rstar/pull/232

Rebasing on top of #238, I get neutral and good results:
```
bulk load baseline      time:   [83.205 µs 83.257 µs 83.311 µs]
                        change: [−3.3614% −3.0514% −2.7456%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 16 outliers among 100 measurements (16.00%)
  1 (1.00%) low severe
  5 (5.00%) low mild
  6 (6.00%) high mild
  4 (4.00%) high severe

insert sequential       time:   [732.40 µs 733.03 µs 733.64 µs]
                        change: [−0.4895% −0.1879% +0.0844%] (p = 0.20 > 0.05)
                        No change in performance detected.
Found 8 outliers among 100 measurements (8.00%)
  3 (3.00%) low mild
  5 (5.00%) high mild

Benchmarking Bulk load complex geo-types geom: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 41.1s, or reduce sample count to 10.
Bulk load complex geo-types geom
                        time:   [408.56 ms 410.21 ms 412.62 ms]
                        change: [−1.0198% −0.4287% +0.2118%] (p = 0.20 > 0.05)
                        No change in performance detected.
Found 12 outliers among 100 measurements (12.00%)
  6 (6.00%) high mild
  6 (6.00%) high severe

Benchmarking Bulk load complex geo-types geom with cached envelope: Collecting 100 samples in estimated 5.1122 s (
Bulk load complex geo-types geom with cached envelope
                        time:   [3.2044 ms 3.2143 ms 3.2265 ms]
                        change: [−2.5921% −1.3668% −0.4515%] (p = 0.01 < 0.05)
Found 7 outliers among 100 measurements (7.00%)
  2 (2.00%) high mild
  5 (5.00%) high severe

bulk load quality       time:   [29.750 µs 29.816 µs 29.901 µs]
                        change: [−15.940% −14.139% −12.365%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 4 outliers among 100 measurements (4.00%)
  3 (3.00%) high mild
  1 (1.00%) high severe

sequential load quality time:   [34.549 µs 35.797 µs 37.286 µs]
                        change: [−16.930% −13.023% −8.8539%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 21 outliers among 100 measurements (21.00%)
  21 (21.00%) high severe

locate_at_point (successful)
                        time:   [239.12 ns 239.56 ns 240.01 ns]
                        change: [−0.7831% −0.5645% −0.3491%] (p = 0.00 < 0.05)

                        time:   [276.25 ns 276.71 ns 277.20 ns]
                        change: [−1.5099% −1.2950% −1.0853%] (p = 0.00 < 0.05)
                        Performance has improved.

locate_at_point_int (successful)
                        time:   [26.113 ns 26.144 ns 26.175 ns]
                        change: [−0.3619% −0.1356% +0.0958%] (p = 0.25 > 0.05)
                        No change in performance detected.
Found 6 outliers among 100 measurements (6.00%)
  3 (3.00%) low mild
  2 (2.00%) high mild
  1 (1.00%) high severe

locate_at_point_int (unsuccessful)
                        time:   [278.05 ns 278.66 ns 279.31 ns]
                        change: [−0.0945% +0.3124% +0.7528%] (p = 0.15 > 0.05)
                        No change in performance detected.
```

rstar/src/point.rs
michaelkirk: I've done this. Note it required explicit implementations (via a macro). Given that we want explict impls for f36/64, I don't think we can do a blanket impl for all of the "other" RTreeNum

rstar/src/aabb.rs
michaelkirk: I pulled this into a separate commit if you want to test it.

michaelkirk:
I believe I've addressed all your feedback. To get this branch building I've rebased it on top of #238. Probably easiest to review and merge those first.

michaelkirk:
Rebased!

rstar/src/point.rs
```diff
@@ -95,9 +156,47 @@ use num_traits::{Bounded, Num, Signed, Zero};
 /// #
 /// ```
 ///
-pub trait RTreeNum: Bounded + Num + Clone + Copy + Signed + PartialOrd + Debug {}
+pub trait RTreeNum: Bounded + Num + Clone + Copy + Signed + PartialOrd + Debug + RTreeCmp {}
```
adamreichold: Can we drop `PartialOrd` here? It would increase our confidence that we did not miss a spot where we should have used `.total_cmp(..)`  instead of `.partial_cmp(..)`.

I also think this is now breaking and we should bump to 0.14 and fix up the change log.

adamreichold (approved):
LGTM with removing the `PartialOrd` bound the only remaining question.

rstar/src/point.rs
michaelkirk: Great point about this being a breaking release. 

At this point, would it make more sense to move total_cmp INTO RTreeNum rather than introducing a separate RTreeCmp trait? 

It would mean our blanket impl on RTreeNum doesn't work, but it's kind of moot because we're now requiring RTreeCmp which similarly has no blanket impl.

michaelkirk: My reason being I don't think we get anything by introducing a second trait. There's nowhere we use RTreeCmp *without* RTree num. Pretending like they are two different things feels overly ceremonious for no tangible gain.

adamreichold: > Pretending like they are two different things feels overly ceremonious for no tangible gain.

Agreed. Ideally, we drop `PartialOrd` and fold `RTreeCmp` into `RTreeNum`.

michaelkirk: Ok great point about dropping partialOrd, it found a bunch of lingering call sites.

Unfortunately, fixing them naively causes a performance regression. I'll have to dig in more next week.

fogti:
If we do this, I think it would be a good idea to also implement the trait for https://docs.rs/rust_decimal/latest/rust_decimal/struct.Decimal.html (which is the only common custom numerical type I know of that does support `Ord`) behind a feature flag because implementing it downstream would be a bad idea (and lead to similar problems as with the `geo` crate).

adamreichold:
> If we do this, I think it would be a good idea to also implement the trait for https://docs.rs/rust_decimal/latest/rust_decimal/struct.Decimal.html (which is the only common custom numerical type I know of that does support `Ord`) behind a feature flag because implementing it downstream would be a bad idea (and lead to similar problems as with the `geo` crate).

I see the problem but would suggest a newtype wrapper downstream for now. (There a derive macros to help with the delegation.)

A proper solution IMHO would be to have `total_cmp` or something like it in `num_traits`. Having an arbitrary order on non-finite values is not specific to R trees after all. (Similarly, I'd still advocate to have a vocabulary types crate for geospatial types on which algorithms like `geo` can depend similar to how all the format crates depend on `serde_core`.)

fogti:
Funnily enough, I recently needed the same: https://github.com/fogti/anyangle/blob/0faa1a847746bc2bc15783c3185be6222749efe7/src/math/mod.rs#L126-L140

So trying to get it into `num-traits` first is probably a good idea.

fogti:
I would have a proposal: https://github.com/fogti/num-traits/commit/92135f1ba4bd6dacd7ca5e64820b65573619b779

but one should alternatively also consider stuff like https://github.com/reem/rust-ordered-float, which would make this complication unnecessary.

adamreichold:
> but one should alternatively also consider stuff like https://github.com/reem/rust-ordered-float, which would make this complication unnecessary.

I don't think these wrapper types would replace a `TotalOrder` trait and also we'd need to choose one which might conflict with downstream choices. I would prefer if you would propose your trait for merging and we could eventually rebase here.

rstar/src/algorithm/nearest_neighbor.rs
```diff
@@ -248,7 +248,7 @@ where
             let distance_if_less_or_equal = match child {
                 RTreeNode::Parent(ref data) => {
                     let distance = data.envelope.distance_2(query_point);
-                    if distance <= *min_max_distance {
+                    if distance.total_cmp(*min_max_distance).is_le() {
```
michaelkirk, line 251: This commit is the result of removing `PartialOrd`.

It ensures we use total_order semantics everywhere, but the new code is much slower than previous. What previously amounted to a couple of machine instructions is now much slower.


e.g. if you pop this into [godbolt](https://godbolt.org):
```rust
fn old(a: f64, b: f64) -> bool {
    a <= b
}

fn new(a: f64, b: f64) -> bool {
 a.total_cmp(&b).is_le()
}

pub fn main() {
    assert_eq!(old(1.0, 2.0), new(1.0, 2.0));    
}
```

```asm
`old`
        movsd   qword ptr [rsp - 16], xmm0
        movsd   qword ptr [rsp - 8], xmm1
        ucomisd xmm1, xmm0
        setae   al
        and     al, 1
        ret

`new`:
        sub     rsp, 24
        movsd   qword ptr [rsp + 8], xmm0
        movsd   qword ptr [rsp + 16], xmm1
        lea     rdi, [rsp + 8]
        lea     rsi, [rsp + 16]
        call    <f64>::total_cmp
        movzx   edi, al
        call    <core[940e698046960fd5]::cmp::Ordering>::is_le
        and     al, 1
        add     rsp, 24
        ret

`<f64>::total_cmp`:
        mov     qword ptr [rsp - 16], rdi
        mov     qword ptr [rsp - 8], rsi
        movsd   xmm0, qword ptr [rdi]
        movq    rax, xmm0
        mov     qword ptr [rsp - 32], rax
        movsd   xmm0, qword ptr [rsi]
        movq    rax, xmm0
        mov     qword ptr [rsp - 24], rax
        mov     rax, qword ptr [rsp - 32]
        sar     rax, 63
        shr     rax
        xor     rax, qword ptr [rsp - 32]
        mov     qword ptr [rsp - 32], rax
        mov     rax, qword ptr [rsp - 24]
        sar     rax, 63
        shr     rax
        xor     rax, qword ptr [rsp - 24]
        mov     qword ptr [rsp - 24], rax
        mov     rax, qword ptr [rsp - 32]
        mov     rcx, qword ptr [rsp - 24]
        sub     rax, rcx
        setl    cl
        setg    al
        sub     al, cl
        ret

<core[940e698046960fd5]::cmp::Ordering>::is_le:
        mov     al, dil
        mov     byte ptr [rsp - 1], al
        cmp     al, 0
        setle   al
        and     al, 1
        ret
```



```
$ cargo bench --bench="*" -- --baseline=main
   Compiling rstar v0.13.0 (/Users/mkirk/src/georust/rstar/rstar)
   Compiling rstar-benches v0.1.1 (/Users/mkirk/src/georust/rstar/rstar-benches)
    Finished `bench` profile [optimized] target(s) in 1.96s
     Running benches/benchmarks.rs (target/release/deps/benchmarks-c2110b35ae6c665a)
Gnuplot not found, using plotters backend
bulk load baseline      time:   [84.828 µs 85.045 µs 85.298 µs]
                        change: [−0.2158% +0.0750% +0.3619%] (p = 0.60 > 0.05)
                        No change in performance detected.
Found 8 outliers among 100 measurements (8.00%)
  7 (7.00%) high mild
  1 (1.00%) high severe

Benchmarking insert sequential: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 5.7s, enable flat sampling, or reduce sample count to 60.
insert sequential       time:   [1.1406 ms 1.1434 ms 1.1464 ms]
                        change: [+58.207% +58.643% +59.061%] (p = 0.00 < 0.05)
                        Performance has regressed.

Benchmarking Bulk load complex geo-types geom: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 128.4s, or reduce sample count to 10.
Bulk load complex geo-types geom
                        time:   [1.2409 s 1.2451 s 1.2495 s]
                        change: [+204.62% +205.63% +206.57%] (p = 0.00 < 0.05)
                        Performance has regressed.

Bulk load complex geo-types geom with cached envelope
                        time:   [3.1861 ms 3.1883 ms 3.1906 ms]
                        change: [−3.5250% −2.5338% −1.6941%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 3 outliers among 100 measurements (3.00%)
  3 (3.00%) high mild

bulk load quality       time:   [53.332 µs 54.685 µs 56.135 µs]
                        change: [+62.018% +66.496% +71.103%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 5 outliers among 100 measurements (5.00%)
  5 (5.00%) high mild

sequential load quality time:   [66.094 µs 67.411 µs 69.152 µs]
                        change: [+60.475% +65.140% +70.040%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 20 outliers among 100 measurements (20.00%)
  1 (1.00%) high mild
  19 (19.00%) high severe

locate_at_point (successful)
                        time:   [248.81 ns 249.07 ns 249.36 ns]
                        change: [+4.6743% +4.8818% +5.0857%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 3 outliers among 100 measurements (3.00%)
  3 (3.00%) high mild

                        time:   [274.65 ns 274.95 ns 275.34 ns]
                        change: [+5.4323% +5.8684% +6.2962%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 15 outliers among 100 measurements (15.00%)
  3 (3.00%) high mild
  12 (12.00%) high severe

locate_at_point_int (successful)
                        time:   [32.153 ns 32.177 ns 32.205 ns]
                        change: [+22.818% +23.025% +23.234%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 8 outliers among 100 measurements (8.00%)
  4 (4.00%) high mild
  4 (4.00%) high severe

locate_at_point_int (unsuccessful)
                        time:   [276.14 ns 276.58 ns 277.03 ns]
                        change: [+6.0457% +6.2393% +6.4298%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 9 outliers among 100 measurements (9.00%)
  5 (5.00%) high mild
  4 (4.00%) high severe
```


Without this commit, we get a perf win and avoid some `--release` crashes.  With this commit we get a perf loss. Personally, I still think this branch is worthwhile *iff* we revert the last commit.

fogti: I think the problem with using `PartialOrd` is that it isn't immediately obvious that using `total_cmp` in some places and `partial_cmp` elsewhere leads to consistent calculations and adheres to the [POLA](https://en.wikipedia.org/wiki/Principle_of_least_astonishment).

michaelkirk: Yes. It is a trade off, and our response is a judgement call. Is it worth the performance hit?

mikwielgus:
Hi, I have a question regarding this PR.

Have you considered the impact of `.total_cmp()` viewing `-0.0` as smaller than `0.0`, not equal? This could create some (literally) edge cases where location queries unexpectedly fail due to bboxes touching only at the origin.

Perhaps `ordered-float`'s approach could be better? It solves this differently; it just views NaNs as always larger than any other value and equal to itself, in contradiction with the IEEE standard.

michaelkirk:
yeesh, that's concerning - thanks for bringing it up.

I'm not super worried about changing around the results of people passing in NaN (garbage) coordinates, because I don't think it's reasonable to expect useful numeric results from NaN, but your example is interesting, because it could arise from completely "valid" input.

I'm not immediately sure what to do about that.

michaelkirk:
> ordered-float's approach could be better?

ordered_float might be interesting... https://docs.rs/ordered-float/latest/src/ordered_float/lib.rs.html#260

It seems like it would only give "unintuitive" results for NaN, and the overhead is an additional `is_nan |` with every comparison which I think should produce much less code than the total_cmp impl above.

mikwielgus:
Just for the record, I should note that, of course, unlike NaNs and infinities, the negative zero in floats is easily obtained from entirely valid operands -- it is the result of multiplication or division of positive zero by any negative number.

adamreichold:
> ordered_float might be interesting... https://docs.rs/ordered-float/latest/src/ordered_float/lib.rs.html#260

We could use that to implement our comparison method in `RTreeNum`. Should probably change the name a bit, e.g. `infallible_cmp` or something. Not sure whether this is still something that could be upstreamed into `num-traits` though.

@michaelkirk Would that also resolve the performance regressions?

Taking a step back, since we are making a breaking change, we could also make modification methods fallible and reject NaN coordinates during insertion/bulk-loading/etc. This would keep this out of the querying paths outright and callers can decide whether to `.unwrap` or not.

urschrei:
> We could use that to implement our comparison method in RTreeNum.

Something like

```rust
impl RTreeNum for f64 {
    #[inline]
    fn infallible_cmp(self, other: Self) -> Ordering {
        match self.partial_cmp(&other) {
            Some(ordering) => ordering,
            None => match (self.is_nan(), other.is_nan()) {
                (true, true) => Ordering::Equal,
                (true, false) => Ordering::Greater,
                _ => Ordering::Less,
            },
        }
    }
}
```

urschrei:
@adamreichold WDYT of the above?

adamreichold:
> @adamreichold WDYT of the above?

Yes, I think that can work, it is just not implemented here yet.

I also think the fallible construction/modification part can work but we should probably only look into if the benchmarks still indicate issues when we use `infallible_cmp` everywhere. (In my personal usage of R trees as spatial indexes, I filter NaN coordinates when I prepare my data for spatial indexing, so I lean towards this personally, but I can reasonable people go either way.)

mikwielgus:
Obviously, one more option is to require the user to pass an `Ord`-implementing type as a generic parameter, and recommend to the user to use a type from `ordered-float` or `typed_floats` crates (`OrderedFloat<...>`, `NonNaN<...>`, `NotNan<...>`).

It will have mediocre ergonomics, sure, but accidental NaNs this are the kind of error static type checking is well-suited to catch, and it avoids potential performance degradation due to imposition of additional runtime NaN checks.

adamreichold:
> It will have mediocre ergonomics, sure, but accidental NaNs this are the kind of error static type checking is well-suited to catch, and it avoids potential performance degradation due to imposition of additional runtime NaN checks.

You can already have that by implementing `RTreeNum` on (your wrapper of) on of these float wrappers, so we are not precluding this, but I wouldn't want this as a default and rather provide something that works with `f32/f64` out of the box as these cover the vast majority of coordinate types as far I know.

adamreichold:
But your proposal does suggest going for `infallible_cmp` as the fallible construction/modification methods would not be necessary for statically checked float wrappers.

michaelkirk:
I will continue to work on this, I've just been distracted the last couple of days.

But if anybody is impatient, I won't be offended if you get there before me.

michaelkirk:
Ok, I pretty much went back to square one on this, so this needs a fresh review. I employed `OrderedFloat`, which I feel produced the "most correct" output with the least overhead. The resulting code is a little ugly (`a.ord() < b.ord()` instead of `a < b`), but it's easy to review and be confident we are using total `Ord` everywhere since `RTreeNum` itself is no longer `PartialOrd`.

re: @urschrei's suggestion:

> Something like
>
> ```rust
> impl RTreeNum for f64 {
>     #[inline]
>     fn infallible_cmp(self, other: Self) -> Ordering {
>         match self.partial_cmp(&other) {
>             Some(ordering) => ordering,
>             None => match (self.is_nan(), other.is_nan()) {
>                 (true, true) => Ordering::Equal,
>                 (true, false) => Ordering::Greater,
>                 _ => Ordering::Less,
>             },
>         }
>     }
> }

This approach is semantically correct, but taken literally, the performance is not tenable (IMO) for the same reasons mentioned in https://github.com/georust/rstar/pull/237#discussion_r3799562515

The ordered_float approach is lighter weight (though still slower than what we have today - that's the cost of doing business with NaN).

The third commit adds support for `NotNan` -  a type checked value (like `NonZero`) that pre-validates numeric types so we can avoid checking for NaN later. Unfortunately it's not that simple: Because we do arithmetic in the crate, we could potentially creating new NaN's (see tests), so there are places within the crate which actually get *more* expensive using `NotNan`. And what's more, to guarantee defined behavior, NotNan will panic if arithmetic creates a NaN, so there's new opportunities for panic if you use `NotNan`.

My (admittedly weak) rationale is that this is just the contract of `NotNan`, and anyone reaching for this sharp knife should be prepared to be cut by it. I'd feel differently if this were some kind of hidden API that rstar used "behind the scenes", but in this case the user has to explicitly add the `ordered_float` crate and instantiate their own `NotNan` themselves - it's not our job to explain the limitations of every external crate. That's my rationale anyway.

The 99% of people who are using f32/f64 can do so without worrying about panics.

Even avoiding `.cmp`, nan-handling is a significant performance regression for some APIs. Here's a summary of the benches:
```
    benchmark                            base    f64 (vs. base)  NotNan (vs. base)
    ---------------------------------- -------- ---------------- ----------------
    locate_at_point_int (successful)    25.42ns  39.96ns (1.57x)  25.41ns (1.00x)
    bulk load baseline                  84.27us  83.64us (0.99x)  83.91us (1.00x)
    locate_within_distance              14.56us  15.45us (1.06x)  14.36us (0.99x)
    locate_at_point (unsuccessful)     277.44ns 279.65ns (1.01x) 275.18ns (0.99x)
    locate_at_point_int (unsuccessful) 277.54ns 280.75ns (1.01x) 275.82ns (0.99x)
    locate_in_envelope_intersecting     17.20us  17.16us (1.00x)  17.33us (1.01x)
    locate_at_point (successful)       233.39ns 231.62ns (0.99x) 238.08ns (1.02x)
    nearest_neighbor_iter, 10 nearest  732.59ns 964.04ns (1.32x) 867.03ns (1.18x)
    insert sequential                  703.29us   1.10ms (1.57x) 943.59us (1.34x)
    nearest_neighbor                    39.55us  54.79us (1.39x)  67.67us (1.71x)
```

Note that `nearest_neighbor` is **slower** for NotNan due to the overhead of arithmetic, but equal or faster  everywhere else.

urschrei:
Should we re-export ordered-float? Or `pub use` it? Otherwise we're coupling rstar consumers who also require it to a compatible version, and version upgrades are a headache for us.

urschrei (approved):

urschrei:
(This is really thorough btw. Thank you.)

rstar/Cargo.toml
fogti: The MSRV needs to be bumped.

michaelkirk: Pushed. Are we all OK to bump to MSRV 1.90?  (released sep 19, 2025)

fogti (approved):

michaelkirk:
I'd like @adamreichold to weigh in too, since this seems like a significant change and he previously expressed some well informed opinions (not trying to rush you, just trying to explain why it's still open after one maintainer's approval).

michaelkirk:
> Should we re-export ordered-float? Or pub use it? Otherwise we're coupling rstar consumers who also require it to a compatible version, and version upgrades are a headache for us.

As I understand it, this would only be relevant to those who want to use `NotNan`, right? The primary usage of OrderedFloat is an internal implementation detail, so doesn't affect the API.

My only hesitation is this paragraph from my previous comment:

> My (admittedly weak) rationale [for being OK with NotNan panic'ing under certain scenarios] is that this is just the contract of NotNan, and anyone reaching for this sharp knife should be prepared to be cut by it. I'd feel differently if this were some kind of hidden API that rstar used "behind the scenes", but in this case the user has to explicitly add the ordered_float crate and instantiate their own NotNan themselves - it's not our job to explain the limitations of every external crate. That's my rationale anyway.

If we re-export NotNan that argument loses some strength.

urschrei:
> The primary usage of OrderedFloat is an internal implementation detail, so doesn't affect the API.

```rust
pub trait RTreeNum: Bounded + Num + Clone + Copy + Signed + Debug {
    type OrdType: Ord;
    fn ord(self) -> Self::OrdType;
}

impl RTreeNum for f64 {
    type OrdType = OrderedFloat<Self>;
    ...
}
```

So I think that makes it public? In practice I don't think it will cause an issue unless you use `NotNan` as you say, but we should still `pub use ordered_float;` so people get the semver version we specify and it can coexist with their version.

PR #238, opened 2026-08-12
fix demo deps

michaelkirk:
---

Edit: note this is based on https://github.com/georust/rstar/pull/232 so merge that first.

`cargo test` / `cargo check` of the workspace currently fails.

Potentially (?) controversial is adding `cargo check` (of the demo) to CI. It could be potentially annoying to have the demo block merging of other work. It's also annoying now that I can't run `cargo check` in the workspace because the demo is broken. 🤷 I'm open to omitting the second commit.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -46,8 +46,11 @@ jobs:
         uses: Swatinem/rust-cache@v2
       - name: Check formatting using Rustfmt
         run: cargo fmt --all --check
-      - name: Check workspace
+      - name: Check main crate
+        run: cargo check --package rstar --all-targets
+      - name: Check bench and demo crates (warn only)
         run: cargo check --workspace --all-targets
+        continue-on-error: true
       - name: Lint using Clippy
         run: cargo clippy --workspace --exclude rstar-demo --tests
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -46,8 +46,6 @@ jobs:
         uses: Swatinem/rust-cache@v2
       - name: Check formatting using Rustfmt
         run: cargo fmt --all --check
-      - name: Check main crate
-        run: cargo check --package rstar --all-targets
       - name: Lint using Clippy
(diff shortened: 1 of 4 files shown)


adamreichold:
> Potentially (?) controversial is adding cargo check (of the demo) to CI. It could be potentially annoying to have the demo block merging of other work. It's also annoying now that I can't run cargo check in the workspace because the demo is broken. 🤷 I'm open to omitting the second commit.

I think we deliberately removed it as to not block things on it. I think a more refined position would be a separate job that builds but does not fail the workflow. This way, we notice something breaking early and automatically, but we still do not block anything on these optional components. WDYT?

michaelkirk:
I've never used these "non-blocking" warning jobs before. It seems kind of useless as is, because there's no indication it failed unless you manually go look at the annotations in https://github.com/georust/rstar/actions/runs/31719804060

Since you brought it up @adamreichold, is there a better way to do this?

adamreichold:
> Since you brought it up @adamreichold, is there a better way to do this?

You added it as a `continue-on-error: true` step to the existing job. I was thinking of a separate job "workspace check" that just is not considered by the `conclusion` job and therefore would not fail the overall build. This should then be visible as a separate "check" in the PR overview.

michaelkirk:
Like this @adamreichold? Otherwise if you have a clear vision of how to do this, please just push something up.

.github/workflows/test.yml
```diff
@@ -46,9 +46,29 @@ jobs:
         uses: Swatinem/rust-cache@v2
       - name: Check formatting using Rustfmt
         run: cargo fmt --all --check
+      - name: Check main crate
```
adamreichold: This step should be redundant to the Clippy step below (Clippy does everything check does and more).

adamreichold (approved):
Indeed like this, just one nit on the check step.

.github/workflows/test.yml
michaelkirk: Good point. Removed.

Issue #197, opened 2025-07-13
RTree::bulk_load does not comply with MAX_SIZE in higher dimensions


AlphaNow37:
This code, using points with a lot of dimensions
```rust
let pts = (0..100_000).map(|i| [i as f32; 100]).collect::<Vec<_>>();
let tree = RTree::bulk_load(pts);
dbg!(tree.root().children().len());
```
prints 100_000 which is far above MAX_SIZE=6

michaelkirk:
Thank you for reporting. 

If I understand correctly, `MAX_SIZE` is intended to be the largest any node can ever be, it seems like a bug in bulk_load_recursive. The node sizes output from that seem to grow exponentially as the number of dimensions in the input increases.

Note this is true even for 2-D points
```rust
   #[test]
    fn test_2d_bulk_load() {
        let pts = (0..25).map(|i| [i as f32; 2]).collect::<Vec<_>>();
        let tree = RTree::bulk_load(pts);
        // > tree.root().children().len() = 9
        dbg!(tree.root().children().len());
        assert!(tree.root().children().len() <= DefaultParams::MAX_SIZE);
    }
```

michaelkirk:
Mostly just some notes at this point: bulk_load uses the [OMT strategy for loading](http://ftp.informatik.rwth-aachen.de/Publications/CEUR-WS/Vol-74/files/FORUM_18.pdf).

(Disclaimer: I'm not a very frequent contributor to this crate,  and it's my first time looking at that paper)

The implementation in the paper is specific to 2-d points.

My understanding is that, the OMT loading strategy should be efficient for querying in higher dimensions too.

I think there are two issues:

1. A simple transcription bug: https://github.com/georust/rstar/blob/4488c142c78e28ee7bc8163f06b26642db5d3958/rstar/src/algorithm/bulk_load/cluster_group_iterator.rs#L66

Should be `floor`, not `ceil`, which is a simple fix, and that change alone resolves the broken [2D behavior I posted upthread](https://github.com/michaelkirk). But not the larger dimensions.

2. The trickier issue, as I understand it (and I'm actually not entirely sure that I do), I think we need to be able to partition along each dimensional axis for OMT to work, which implies that the minimum node size must be at least `2^num_dimensions`. Which further implies that OMT is not a suitable algorithm for high dimensional data.

Can anyone more knowledgable weigh in?

PR #239, opened 2026-08-25
fix(bulk_load): keep all leaves on the same depth

### LINKED ISSUE #45: Unreachable! panic when inserting points
I'm seeing unreachable! panics when I insert certain configurations of points. An example which reproducibly fails for me is here: https://gist.github.com/alec-deason/6218fb2fd6262bdd63473e417e050004

I've stripped that down so that changing the position of any of the last few points even slightly will flip it from failing to succeeding. Adding or removing any points either from the bulk loaded set or the ones that are inserted later makes the failure go away. Also inserting all of he points one at a time rather than bulk loading the initial set causes the failure to go away in this example, though I haven't experimented enough to see if the failure only appears with bulk loading.

The program that generated this list of points hits this failure case regularly but not with every set it generates. I have seen it hit the unreachable! panic in both algorithm::rstar::forced_insertion and algorithm::rstar::recursive_insert, which look to be the same underlying failure.

I'm using nightly. I see the same behavior with rstar 0.8.1 and with the latest commits.

jaideeppyne:
`RTree::bulk_load` built the tree by recursively partitioning elements and turning any cluster with ≤ `MAX_SIZE` elements into a leaf. When the element count made sibling clusters straddle `MAX_SIZE` (some just below, some just above), the larger clusters were split one level deeper than the smaller ones, so leaves ended up on **different depths**.

Such a tree violates the R-tree invariant that all leaves share the same depth (the invariant `ParentNode::sanity_check` asserts). Queries still worked, but a later `insert` descends assuming uniform depth and panics with `"This is a bug in rstar."` on reaching a node that mixes leaf and parent children. Reproducible with default parameters in 2-D, e.g. bulk-loading 25 points and then inserting two more:

```rust
let mut t = RTree::bulk_load(vec![[12,-8],[-8,-9], /* … 25 points … */]);
t.insert([-8,-1]);
t.insert([3,9]); // panicked
```

Fix: compute the target tree depth up front and thread it through the recursion so every branch stops on the same level. Adds a regression test and a CHANGELOG entry. Full suite + fmt + clippy pass.

This is distinct from #197 (which is about `MAX_SIZE` compliance in higher dimensions — the repro here keeps every child count within `MAX_SIZE`) and from the long-closed #45 (a reinsertion bug; here the tree is already malformed straight out of `bulk_load`, before any insert). It does not attempt to fully solve #197.

- [x] I agree to follow the project's Code of Conduct.

### DIFF
--- a/rstar/src/algorithm/bulk_load/bulk_load_sequential.rs
+++ b/rstar/src/algorithm/bulk_load/bulk_load_sequential.rs
@@ -12,14 +12,54 @@ use num_traits::Float;
 
 use super::cluster_group_iterator::{calculate_number_of_clusters_on_axis, ClusterGroupIterator};
 
-fn bulk_load_recursive<T, Params>(mut elements: Vec<T>) -> ParentNode<T>
+/// Computes the depth (number of node levels, leaves included) of the tree that
+/// bulk loading will build for `number_of_elements` elements.
+///
+/// This mirrors the partitioning performed by [`bulk_load_recursive`]: on every
+/// level a node is split into `number_of_clusters_on_axis ^ DIMENSIONS`
+/// clusters, so a balanced cluster shrinks by that factor per level. The
+/// returned depth is the first level at which a balanced cluster fits into a
+/// single (leaf) node, i.e. holds at most `MAX_SIZE` elements.
+///
+/// Threading this depth through the recursion (instead of stopping each branch
+/// individually as soon as it reaches `MAX_SIZE` elements) guarantees that all
+/// leaves end up on the same level. Otherwise clusters whose size straddles
+/// `MAX_SIZE` -- some just below, some just above -- would produce leaves on
+/// different levels, yielding a malformed R-tree that violates the "all leaves
+/// share the same depth" invariant and makes a subsequent `insert` panic.
+fn bulk_load_depth<T, Params>(number_of_elements: usize) -> usize
+where
+    T: RTreeObject,
+    Params: RTreeParams,
+{
+    let m = Params::MAX_SIZE;
+    let dimensions = <T::Envelope as Envelope>::Point::DIMENSIONS;
+    let mut depth = 1;
+    let mut cluster_size = number_of_elements;
+    while cluster_size > m {
+        let number_of_clusters_on_axis =
+            calculate_number_of_clusters_on_axis::<T, Params>(cluster_size).max(2);
+        // Number of clusters this node is split into (across all axes).
+        let mut fanout = 1usize;
+        for _ in 0..dimensions {
+            fanout = fanout.saturating_mul(number_of_clusters_on_axis);
+        }
+        // Size of a balanced cluster on the next level. Rounding down selects
+        // the shallowest depth that keeps clusters from underflowing, which
+        // avoids creating nodes with fewer than `MIN_SIZE` children.
+        cluster_size /= fanout.max(2);
+        depth += 1;
+    }
+    depth
+}
+
+fn bulk_load_recursive<T, Params>(mut elements: Vec<T>, remaining_depth: usize) -> ParentNode<T>
 where
     T: RTreeObject,
     <T::Envelope as Envelope>::Point: Point,
     Params: RTreeParams,
 {
-    let m = Params::MAX_SIZE;
-    if elements.len() <= m {
+    if remaining_depth <= 1 {
         // Reached leaf level. Shrink excess capacity so the in-place collect
         // (which reuses the allocation when size_of::<T> == size_of::<RTreeNode<T>>)
         // doesn't preserve a massively over-sized buffer in the final tree node.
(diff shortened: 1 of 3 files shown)


rstar/src/algorithm/bulk_load/bulk_load_sequential.rs
```diff
@@ -12,14 +12,54 @@ use num_traits::Float;
...
-    <T::Envelope as Envelope>::Point: Point,
     Params: RTreeParams,
 {
     let m = Params::MAX_SIZE;
-    if elements.len() <= m {
+    let dimensions = <T::Envelope as Envelope>::Point::DIMENSIONS;
+    let mut depth = 1;
+    let mut cluster_size = number_of_elements;
+    while cluster_size > m {
+        let number_of_clusters_on_axis =
```
adamreichold, line 40: While I think I can follow this approach, I think it does recompute the number of clusters redundantly (once during depth computation and then again during the actual partitioning).

Couldn't we fix this as well by changing `ClusterGroupIterator` to actually produce two `MAX_SIZE`-sized clusters in the edge case? Probably by counting clusters instead of repeatedly substracting `slab_size` from `self.remaining.len()`?

fallenmi (approved):
Reviewed by OpenAI Codex for @fallenmi. No repository files were changed on GitHub. I verified exact head 32367e5f9e8ca0ea884c3323890ed5b6820d3918 against exact base afae2ae6db9ceb234454812c54be2121b0f90a66. An identical external public API regression with the reported 25 points and two inserts panics on the base at rstar/src/algorithm/rstar.rs:137 with "This is a bug in rstar.", while the head completes with size 27. On the head, 963 deterministic bulk-load and insert cases across sizes 0 through 320 and 2D, 3D, and 4D passed, along with 52 unit tests, 27 documentation tests, and the two external oracle tests. git diff --check is clean and all six live checks succeed. The common remaining depth removes the mixed leaf depth that made insertion hit the impossible descent branch. Approved.

jaideeppyne:
Hold off on merging this for a moment, I found a regression in it.

The depth calculation I added divides the cluster size by the combined fan-out and rounds down, but `ClusterGroupIterator` cuts a slab of `len.div_ceil(number_of_clusters_on_axis)` once per axis. So it understates the largest cluster on the next level, stops one level early, and leaves nodes above `MAX_SIZE`.

It shows up in 2D, where master is clean. Smallest case is 25 collinear points:

```rust
let pts: Vec<[f64; 2]> = (0..25).map(|i| [i as f64, i as f64]).collect();
let tree = RTree::<[f64; 2], DefaultParams>::bulk_load_with_params(pts);
// a node ends up with 7 children, MAX_SIZE is 6
```

Sweeping n = 1..300 on collinear input, master has no violations and this branch has them from n = 25. My own test here doesn't catch it because it calls `sanity_check::<DefaultParams>(false)`, and `false` is precisely the flag that skips the `MAX_SIZE` assertion. With `true` it panics at n = 25.

Rounding the depth up per axis instead is not enough on its own: it pushes some trees a level deeper and then nodes fall below `MIN_SIZE`, which is what the rounding-down comment in the patch was guarding against.

@adamreichold you suggested changing `ClusterGroupIterator` rather than computing the cluster count twice, and I think you were right, just for a bigger reason than the redundancy. Fixing it in the iterator is what actually resolves this.

What does work is fixing the fan-out itself. Splitting every axis by the same number gives a node `clusters ^ DIMENSIONS` children, which is already over `MAX_SIZE` as soon as there is more than one axis to split, so 3D has 8-child nodes and 6D has 64-child nodes on master today. That is #197. If instead the cuts are handed out one axis at a time, always to the axis cut least so far, and only while the product stays inside `MAX_SIZE`, then the largest node is 6 in 2D through 6D and leaf depth stays uniform. Full suite passes, so `MIN_SIZE` holds too.

I have that on a branch. Happy to either fold it into this PR, which would mean you re-reviewing since it touches the code this PR adds, or land this one first with just the depth rounding sorted and do the fan-out separately as a fix for #197. Your call on which is less annoying to review.

One thing worth flagging if we go the fan-out route: it makes 4D and up substantially faster (6D nearest-neighbour roughly halves) but 3D range queries get slower, because cutting all three axes in two needs a fan-out of 8 and `MAX_SIZE` is 6, so 3D loses a split that the oversized nodes were getting for free. That also answers @michaelkirk's question in #197 about splitting every axis: it is not compatible with `MAX_SIZE` < 2^DIMENSIONS.

I used an LLM to build the differential harness and the sweeps behind this, as with the original PR.

rstar/src/algorithm/bulk_load/bulk_load_sequential.rs
jaideeppyne: Coming back to this: you were right that the iterator is the place to fix it, and it turns out to matter more than the redundant recomputation. The depth calculation as written rounds down where the iterator rounds up per axis, so it stops a level early and this branch leaves 7-child nodes in 2D, which master does not. Details and the numbers are in a top level comment just now.

PR #241, opened 2026-09-01
RTree/serde: Fix `UnsupportedType(Some(PhantomData))` on toml serialization

fogti:
---

## Context

TOML doesn't support `PhantomData`, and I ran into this while testing https://github.com/mikwielgus/anyangle/pull/45.


adamreichold (approved):
Thanks!

PR #246, opened 2026-09-07
Fix overflow panic draining an empty tree with integer coordinates

MaxFreedomPollard:
- [x] I agree to follow the project's [code of conduct](https://github.com/georust/.github/blob/main/CODE_OF_CONDUCT.md).
- [x] I added an entry to the project's change log file if knowledge of this change could be valuable to users.
---

`drain_within_distance` panicked with an arithmetic overflow on an empty tree with integer coordinates, because the removal function was handed the empty root envelope whose corners are the scalar type's maximum and minimum. The drain iterator now seeds its stack only when the root has children, matching the guard `SelectionIterator::new` already has. `test_drain_within_distance_on_empty_tree` covers a tree that was never filled and one emptied by removals, and it fails with the overflow if the guard is removed.

### DIFF
--- a/rstar/src/algorithm/removal.rs
+++ b/rstar/src/algorithm/removal.rs
@@ -100,7 +100,12 @@ where
         let m = Params::MIN_SIZE;
         let max_depth = (original_size as f32).log(m.max(2) as f32).ceil() as usize;
         let mut node_stack = Vec::with_capacity(max_depth);
-        node_stack.push((root, 0, 0));
+
+        // Do not push on an empty root onto `node_stack` as
+        // its AABB is pathological and might make `should_unpack_parent` panic.
+        if !root.children.is_empty() {
+            node_stack.push((root, 0, 0));
+        }
 
         DrainIterator {
             node_stack,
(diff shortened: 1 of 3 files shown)


adamreichold:
This should not be necessary during every iteration step but only during iterator construction as the pathological AABB with only be forced upon the user by us for an empty tree as described in #184, in any other case the supplied coordinates were pathological to begin with.

Also please don't post uselessly verbose LLM-generated PR descriptions. Everything that is meant to communicate with humans like us maintainers should be written by them.

MaxFreedomPollard:
I'm sorry it's verbose, noted for this repo.

rstar/src/algorithm/removal.rs
```diff
@@ -100,7 +100,16 @@ where
...
         let max_depth = (original_size as f32).log(m.max(2) as f32).ceil() as usize;
         let mut node_stack = Vec::with_capacity(max_depth);
-        node_stack.push((root, 0, 0));
+        // An empty root is never put on the stack, so the removal function never
+        // sees it. It carries the empty envelope, whose corners are the scalar
+        // type's maximum and minimum value, and measuring a distance to that
+        // overflows for integer coordinates. `SelectionIterator::new` guards the
+        // same way. Leaving the stack empty ends the iteration right away and
+        // keeps the empty root that was just swapped in, which is what the tree
+        // held anyway.
```
michaelkirk: It's hard to understand code when it's interspersed with paragraphs of narrative. I'm assuming this was claude generated. Try to put human effort into succinct comments for human consumption.

michaelkirk (approved):
Fix LGTM, thanks! 

I have a non-blocking concern about the documentation, but happy to merge either way.

rstar/src/algorithm/removal.rs
adamreichold: I had this exact same concern but was unsure whether to voice it as it might be considered unreasonable. So thank you for voicing this! I'll rebase and amend the commits myself and then merge this...

(`<rant>`We used to ask people to comment this stuff they wouldn't do so. Similarly with the commit messages and PR descriptions. Remember the days of "fix things"? Now we get spammed with pages and pages of prose that actually says very little. Why can't we have nice things?`</rant>`)

adamreichold (approved):

PR #247, opened 2026-09-08
Remove `RTreeObject` trait bound on struct def. via default generic param.

mikwielgus:
Bounds on struct definitions, unlike bounds on functions, are usually unidiomatic in Rust, and it is usually possible to rewrite the code so that there are only bounds on functions, and nothing on the type definition itself. If you care about this, it becomes a contagious problem to rely on `rstar` in your own generic code, because the `T: RTreeObject` bound propagates upward.

For example,

```rust
pub struct Layout<T: RTreeObject> {
    objects: Vec<T>,
    rtree: RTree<T>,
}
```

must have an `RTreeObject` bound on `T`, even though this code's authors may not want to expose that to their downstream in their code's public interface.

(the above of course also applies to enums)

See also this Stack Overflow answer: https://stackoverflow.com/a/66369912

This PR is a proposal to solve this problem by adding an additional generic parameter `E` to `rstar`'s types that defaults to `RTreeObject::Envelope`.

---

### DIFF
--- a/rstar/src/node.rs
+++ b/rstar/src/node.rs
@@ -13,21 +13,18 @@ use serde::{Deserialize, Serialize};
 #[cfg_attr(
     feature = "serde",
     serde(bound(
-        serialize = "T: Serialize, T::Envelope: Serialize",
-        deserialize = "T: Deserialize<'de>, T::Envelope: Deserialize<'de>"
+        serialize = "T: Serialize, E: Serialize",
+        deserialize = "T: Deserialize<'de>, E: Deserialize<'de>"
     ))
 )]
 /// An internal tree node.
 ///
 /// For most applications, using this type should not be required.
-pub enum RTreeNode<T>
-where
-    T: RTreeObject,
-{
+pub enum RTreeNode<T, E = <T as RTreeObject>::Envelope> {
     /// A leaf node, only containing the r-tree object
     Leaf(T),
     /// A parent node containing several child nodes
-    Parent(ParentNode<T>),
+    Parent(ParentNode<T, E>),
 }
 
 /// Represents an internal parent node.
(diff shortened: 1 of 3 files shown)


rstar/src/node.rs
```diff
@@ -13,21 +13,18 @@ use serde::{Deserialize, Serialize};
 #[cfg_attr(
     feature = "serde",
     serde(bound(
-        serialize = "T: Serialize, T::Envelope: Serialize",
-        deserialize = "T: Deserialize<'de>, T::Envelope: Deserialize<'de>"
+        serialize = "T: Serialize, E: Serialize",
```
fogti, line 16: These special bounds can afaik be removed now, as they are equal to the default

rstar/src/rtree.rs
```diff
@@ -173,16 +174,15 @@ where
 #[cfg_attr(
     feature = "serde",
     serde(bound(
-        serialize = "T: Serialize, T::Envelope: Serialize",
-        deserialize = "T: Deserialize<'de>, T::Envelope: Deserialize<'de>"
+        serialize = "T: Serialize, E: Serialize",
```
fogti, line 177: Same here.

fogti (changes requested):

rstar/src/rtree.rs
```diff
@@ -173,16 +174,15 @@ where
...
+        deserialize = "T: Deserialize<'de>, E: Deserialize<'de>"
     ))
 )]
-pub struct RTree<T, Params = DefaultParams>
+pub struct RTree<T, Params = DefaultParams, E = <T as RTreeObject>::Envelope>
```
adamreichold, line 181: Adding a generic type parameter that is not really a free parameter does not make sense IMHO. It suggests that `RTree<T, Params, E>` means something even if `E = <T as RTreeObject>::Envelope` does not hold and this just is not the case. Or at least it is definitely not something I would want to support in the public API of this crate.

Meaning this might be a reasonable trick for the implementation details of a crate, but not for its public API.

